PR #39, opened 2020-06-18
Update yarn_install task to include modules for EventStreamServer.js.

jsoules:
PR adds additional yarn call to install dependencies for EventStreamServer.


magland:
Looks good. Merging.

PR #42, opened 2020-06-22
Separate plot labels from plots. Improve selection highlights.

jsoules:
This PR does a fair amount of polishing for plot display:

 - Plot title no longer in the plot, but instead rendered in an HTML div component by the plot generator
 - Added separate styling for the focused element, so it's clear what shift-select will do (this also better matches what mountainview does)
 - Added an 'unselected' styling with a transparent border so that the elements inside the boxes don't move around when you click them
 - Ensured that deselection trumps the focus styling
 - Centered the 'loading' spinner inside the plots

Functionality is also improved:
 - The list of unit IDs is now displayed in clickable spans, with shared state between the selected unit IDs (visible from styling) and the selected plots. Clicking either one (which follow the same logic) will modify the state of the other.
 - To facilitate this, I bumped the selection logic, and the logic for identifying whether an individual component is selected/focused, up to the `SortingView` object. It is now passed in and can be shared.
 - Note this means some of the selection and selection-hook data is no longer being passed to individual components. I figure it's easy enough to add it back in, if we should wind up later on implementing components that need to do so.
 - Plot highlight/focus styling rules are now part of an object that's passed in to individual plot components, so they can be defined centrally and shared consistently.

Ongoing Todos:
 - There's still no logic calculating plot size (it's still hard-coded in pyplot, so it's still hard-coded in all the sizing styling which I now have living in `SortingView.js`).
 - There might be a more optimal way to organize the code that's now found a home in `SortingView.js`.

### DIFF
--- a/src/components/MatplotlibPlot.js
+++ b/src/components/MatplotlibPlot.js
@@ -2,6 +2,7 @@ import React, { useState, useEffect, useRef } from 'react'
 import { createHitherJob } from '../hither'
 import { sleep } from '../actions';
 import { CircularProgress } from '@material-ui/core';
+import { Box } from '@material-ui/core';
 const mpld3 = require('./mpld3.v0.3.js');
 
 const MatplotlibPlot = ({ functionName, functionArgs }) => {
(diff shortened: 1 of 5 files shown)


magland:
Styling looks a lot better!
One issue. If you Ctrl+Click to unselect a unit, the highlighting stays on -- I think because the unit becomes the focused unit. I think following should be enforced: the focused unit must always be among the selected units.

jsoules:
Yeah. Keeping a focus despite it not being selected is a feature of how ctrl-selection works (at least in my file browser, which is the model I was using here), but I just pushed a change to reverse the order of the comparisons--now a deselected style will take precedence over the focus style for non-selected elements.

magland:
Tested and it works perfectly. Merging.

PR #43, opened 2020-06-24
WIP - making placeholders for new sorting view plugin widgets

magland:


magland:
To be consistent with other parts of the code, I changed the prop name from handleUnitClicked to onUnitClicked. The name of the handler is still handleUnitClicked, but the callback name is onUnitClicked.

magland:
Started new sorting view: Units. This will be a table of units with info such as unit id, avg. firing rate, num. events, snr, etc.

A few todo items are in the code... both in Units.js and in NiceTable.js.

Fixed the Autocorrelograms prototype data, so that should now show up properly in the prototypes page. (Although data does need to be downloaded already for that example sorting).

For convenience during development: Replaced `isSelected(unitId)` by `(isSelected && isSelected(unitId))` so that the prop isSelected can be undefined.


Moved plotStyles to its own plotStyles.js file rather than passing it in as a prop. I think it's better to do it this way.

magland:
Added an ExampleSortingView that serves as a boilerplate for creating new sorting views. Also separated out sampleSortingViewProps for use with the prototype page.

PR #52, opened 2020-07-23
misc updates to allow loading static feed

magland:
After you pull this, you should be able to view the timeseries for the static feed at:
http://localhost:<PORT>/default/timeseriesForRecording/cortexlab-single-phase-3-ch0-7.10sec?feed=sha1://0ac52651c5b332cc2058cdaa7675d9da8c892669/feed.json


jsoules:
I can't comment on the Bin1Extractor stuff, but everything else looks good. I pulled the branch and checked the performance, which seems nice. I did run with an eye on the network activity and confirmed that chunks are loaded by kachery as needed. I think we're ready to merge this, so I will do so.

PR #53, opened 2020-07-23
Tweaks to firing rate display. Updated renamed React component.

jsoules:
Tweaked names of two columns in firing rate data and rounded to 2 decimal places (ensuring that trailing zeroes are included). The latter also involved doing the division as Decimal and quantizing the result, so I broke up the map function line spacing a little bit so it's easier to read.

As a drive-by, I also got a warning that React's `ExpansionPanel` element family has been renamed to `Accordion`, so I made that update in the codebase.


jsoules:
Wanted to bump this since it's the only outstanding manual PR.

PR #61, opened 2020-08-19
Units rendering

jsoules:
This PR comprises a number of changes intended to improve the efficiency of the UI, specifically around updates to selection status in the units sidebar on the left-hand side of the Sortings View. Additionally, it adds a layer of abstraction for the display of plots such as the average waveforms and the autocorrelograms, so that the mechanics of plotting individual plots can be separated from laying them out and attaching unit-selection functionality and selection status highlights. This abstraction layer is represented in `src/components/PlotGrid.js`.

To improve unit highlight performance, the following changes were made:
* Broke out individual rows of the sidebar into their own component (`SortingInfoUnitEntry.js`)
* Ensured that this component is wrapped in `React.memo` so that it will not be considered for rerendering when its props do not update
* Collected various applied statuses into actual CSS classes located at the root (`localStyles.css`), and applied statuses to components by manipulating CSS class rather than inserting styling information (`SortingInfoUnitEntry.js`, `Autocorrelograms.js`, `AverageWaveforms.js`)
* Replacing callbacks used to identify units' selected/focused status (from global state) with actually passing the list of selected/focused units into the various PluginComponents.
* Replaced most `useState`-based state management in `SortingView.js` with `useReducer` to ensure functions would not need to be re-created with each render (which in turn forced a rerender of all downstream components using those functions as a callback). This resulted in the largest gains. As part of this, I removed the special 'focused' status and simply added a 'focused' member entry to the state object.
* Necessary follow-on changes to `Units.js`, `CrossCorrelogram.js` due to removal of callbacks & changed structure for state management
* Removed extraneous console.log entry from `AverageWaveform.js`.

Outstanding:
--
* Restructuring the `Units` table to improve performance/reduce refreshes and to allow cells to be handled more individually (which will enable partial rendering of rows independent of columns, rather than the current row-at-once, table-complete-before-rendering approach)
* Evaluating performance of different components in combination, since page performance changes for different components when they are displayed in different combinations (possibly due to shared state?)


magland:
This is excellent. I successfully merged these modifications into the neuropixels branch and things worked well there. So I will merge this into master.

PR #62, opened 2020-08-20
change up the client/server communication

magland:


src/hither/createHitherJob.js
```diff
@@ -1,216 +1,210 @@
...
+    errorString: msg.error_message,
+    runtime_info: msg.runtime_info
+  })
+  dispatchUpdateHitherJob({jobId: job._jobId, update: job.object()});
+}
+
 const dispatchAddHitherJob = (job) => {
-  if (!globalDispatch.dispatch) return;
-  globalDispatch.dispatch(addHitherJob(job));
+  if (!globalData.dispatch) return;
```
jsoules, line 69: I reckon this is probably a "can't-happen" sort of thing, but should it throw an error if it does occur, or is it expected?

PR #67, opened 2020-08-28
Add isi violations metric.

jsoules:
Confirmed that we can use additional kwargs through a configuration object in the individual metric plugin. Fixed a few bugs in the sentinel values that were causing display problems/bad logic when the value to be displayed actually was 0.

Confirmed that values are displayed when this is run against an actual dataset (used allen 419112 from the spike workshop).

Possible TODO: We may want to set a maximum decimal precision on the output, so it looks cleaner.

### DIFF
--- a/src/pluginComponents/Units/Units.js
+++ b/src/pluginComponents/Units/Units.js
@@ -61,7 +61,8 @@ const Units = ({ sorting, recording, selectedUnitIds, extensionsConfig,
         return 0;
     });
 
-    const fetchMetric = useCallback(async (metric = {metricName: '', hitherFnName: '', hitherConfig: {}}) => {
+    const fetchMetric = useCallback(async (metric = {metricName: '', hitherFnName: '',
+                                                    metricFnParams: {}, hitherConfig: {}}) => {
         const name = metric.metricName;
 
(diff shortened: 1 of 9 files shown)


magland:
Looks great! I think we can do 3 decimal places. `toFixed(3)`

magland:
Also, here's an idea. We will probably want to reduce the horizontal width of the metrics column labels (as we get more metrics in the system). So let's change the column label to:

`ISI viol.`

And we should specify the tooltip as `ISI violation rate` (notice the lowercase "v" here)

Tooltips can be specified using the "title" prop, as follows:

```
                pluginLabels.map(plugin => (
                        <TableCell key={plugin + '_header'}>
                            <span title={plugin.tooltip}>{plugin.label}</span>
                        </TableCell>
                    ))
```

magland:
... follow up from last comment...

So we would have something like:

```
IsiViolations.metricName = 'IsiViolations';
IsiViolations.columnLabel = 'ISI viol.';
IsiViolations.tooltip = 'ISI violation rate';
IsiViolations.hitherFnName = 'get_isi_violation_rates';
...
```

magland:
Merging.

PR #68, opened 2020-08-31
created new appbar and added dark theme

LastingDynamics2:
- Created a new app bar in `src/components/RootAppBar/RootAppBar.js` and replaced with the old one

- added reducer and actions for toggle dark mode

- added `/utils/featureFlags` module where we store the flags for turning on/off parts of UI


magland:
The new app bar looks great!

Since we use yarn, I would prefer that we not track the package-lock.json file.

I guess we'll need to figure out what to do about the DBC logo. Obviously that should only be there for the DBC version. On the other hand we should try not to have too much divergence between the versions. I guess there are a few options. I'm sort of in favor of having a flat that turns on/off the DBC logo. I guess it's something to discuss via email.

image

jsoules:
Hi! Thanks for your recent PRs. It's not clear whether these are ready for review/merging. We're happy to look at code if there's something you want to share before it's in a merge-ready state. But if that's the case, could I ask you to use a draft PR instead of a full pull request? I think this is really useful whenever you want to share something to keep everyone on the same page or to ask for feedback, but you still expect to make more changes.

When this (and the other PR #70, and future PRs) are actually ready for merge, I'd ask that you do a couple things:
 * Please call us out to make sure we know you're waiting on a review
 * Please give us some detail about what the PR is trying to do and what changes you made to accomplish that (as well as any fixes or code cleanup you made that aren't directly related to the main purpose). For instance, the commit message here gives a good idea of the main goal of the PR, but it'd be nice to include a bullet list of what changed about each of the 11 files touched, so we can know what we're looking at. Either a note per file, or list big-picture changes and the files affected--just something really simple like:
    * `src/index.js`: Added memoized function to get selected theme type; added theme selection to data store; added top-level `App` component
Or
    * Moved toolbar from `src/AppContainer.js` into its own file in `src/components/RootAppBar/RootAppBar.js`.
I've just found this helps to understand a PR, especially when they have larger scope.
  * Finally, since we're working in a project that doesn't yet have thorough test coverage, it'd be good to mention what you think the changes could affect and what you've done to make sure your code does what you intended and doesn't break anything.

Thanks and we look forward to working with you. Just let us know when you're ready to move forward with these and we'll give a review.

magland:
Closing for now. Will work on merge process later on.

PR #72, opened 2020-09-01
split genplot_average_waveforms into 2 steps

magland:
split genplot_average_waveforms into 2 steps

I split the genplot_average_waveforms into a preprocessing step (creating an .hdf5 file with a subsampling of snippets) and a computation script. The idea is that this preprocessing is also used by other metrics. In fact, the same preprocessing job is used by each average_waveform job across all of the units. Hither handles this nicely so that the job is only queued once.

TODO: the creating of the snippets file needs to made more efficient for large problems (long recordings, many channels, many units, many spikes) -- need to look at the innards of the spiketoolkit implementation

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,10 +52,10 @@ Then:
 
 ```bash
 # On Linux:
-labbox-launcher run magland/labbox-ephys:0.3.14 --docker_run_opts "--net host" --data $LABBOX_EPHYS_DATA_DIR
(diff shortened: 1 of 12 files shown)


jsoules:
This all seems perfectly reasonable to me...

magland:
Great! Once I finish optimizing the prepare_snippets_h5 function for large datasets and test, I will turn this into a non-draft PR.

PR #82, opened 2020-09-03
Fix correlogram axis values.

jsoules:
This PR fixes the base axis on the auto/cross correlograms.

After chasing my tail for a bit, I realized that the issue with average waveforms is more complicated: essentially we can't just convert the base unit, because we might actually need to interpolate new values if we're trying to upscale. (Or conversely, how do we go about choosing the downscaling?) I'll look into this more tomorrow, but I wanted to get this PR up ASAP since I don't want our messaging to be stuck waiting on me fixing the other remaining issue.

The other issue is that this PR apparently cannot be automatically merged; I'm not sure what the conflict is, but half of the `src/pluginComponents/AverageWaveforms/genplot_average_waveform.py` file shows missing functions when I try to merge against master. I will fix this tomorrow if you're willing to wait.


magland:
Thx, I resolved the conflict and am merging.

PR #87, opened 2020-09-04
Convert frames to msec for spike waveforms.

jsoules:
PR converts frame counts to milliseconds for the average spike waveform view. Spikes are now in the appropriate ~2ms range.

Within `AverageWaveform_ReactVis.js`, we had just been using the index of each frame entry as the x-coordinate for the plot. I converted these to msec by multiplying by 1000/sampling_frequency (where sampling frequency is in Hz).

Adjusted `fetch_average_waveform_plot_data` in `genplot_average_waveform.py` to pass along the sampling frequency as part of the plot data so that it would be available in `AverageWaveform_ReactVis.js`.

### DIFF
--- a/src/pluginComponents/AverageWaveforms/genplot_average_waveform.py
+++ b/src/pluginComponents/AverageWaveforms/genplot_average_waveform.py
@@ -6,7 +6,7 @@ import spikeextractors as se
 import spiketoolkit as st
 from labbox_ephys import prepare_snippets_h5
 
-@hi.function('createjob_fetch_average_waveform_plot_data', '')
+@hi.function('createjob_fetch_average_waveform_plot_data', '0.1.0')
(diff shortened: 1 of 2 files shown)


src/pluginComponents/AverageWaveforms/genplot_average_waveform.py
```diff
@@ -42,6 +42,7 @@ def fetch_average_waveform_plot_data(snippets_h5, unit_id):
 
     return dict(
         channel_id=int(maxchan_id),
+        sampling_frequency=sampling_frequency.item(),
```
magland, line 45: I would expect sampling_frequency to be a number here. Why `.item()`?

magland (changes requested):
I'm just wondering why .item() is needed in sampling_frequency.item()

magland (approved):

PR #106, opened 2020-10-07
Update README.md

khl02007:
changed the name of example data directory to match the environmental variable


jsoules:
I'm not sure this is still relevant to the current state of the readme. @khl02007, can you confirm whether the problem you were correcting is still present in the current version? Thanks!

khl02007:
You're right - this is no longer relevant.

PR #109, opened 2020-10-16
AverageWaveformsPlotNew

magland:
"Average waveforms new" now plots waveforms at electrode locations


jsoules:
Has this also been superseded, or is this code still outstanding?

magland:
I think it may need to be merged in, but probably by hand because canvas layout interface may have changed.

PR #114, opened 2020-11-24
Canvas jfm

magland:


src/components/jscommon/CanvasWidget/CanvasWidgetNew.tsx
```diff
@@ -83,41 +83,46 @@ interface Props<T extends BaseLayerProps> {
...
-        if (!state.dragRect) return
-        const pixelDragRect = {
-            xmin: state.dragRect[0],
-            xmax: state.dragRect[0] + state.dragRect[2],
-            ymin: state.dragRect[1] + state.dragRect[3],
-            ymax: state.dragRect[1]
-        }
-        for (let l of props.layers) {
-            l.handleDrag(pixelDragRect, released, state.shift, state.dragAnchor, state.dragPosition)
+        if ((dragState) && (dragState !== prevDragState) && (dragState.dragRect)) {
```
jsoules, line 109: Do we need to do a deep compare on this? Or is the state hook making it effectively immutable/you'd wind up replacing the object if any of the properties are changed?

src/components/jscommon/CanvasWidget/CanvasWidgetNew.tsx
```diff
@@ -83,41 +83,46 @@ interface Props<T extends BaseLayerProps> {
...
+            }
+            for (let l of props.layers) {
+                l.handleDrag(pixelDragRect, !dragState.dragging, dragState.shift, dragState.dragAnchor, dragState.dragPosition)
+            }
         }
-    }, [props.layers, state.dragAnchor, state.dragRect, state.dragPosition, state.shift])
+        setPrevDivElement(divElement)
+        setPrevDragState(dragState)
+        setPrevWidth(props.widgetProps.width)
+        setPrevHeight(props.widgetProps.height)
```
jsoules, line 124: As you'd suggested, I was hoping to avoid having to do manual prior-state comparisons, but I guess they're still necessary... bummer. Is it because React is doing a shallow comparison on the `props` dependency, so it winds up 'changing' even if it doesn't actually change?

src/components/jscommon/CanvasWidget/CanvasWidgetNew.tsx
```diff
@@ -127,26 +132,21 @@ const CanvasWidget = <T extends BaseLayerProps>(props: Props<T>) => {
 
     const _handleMouseMove = useCallback((e: React.MouseEvent<HTMLCanvasElement, MouseEvent>) => {
         const { point, mouseButton, modifiers } = ClickEventFromMouseEvent(e, ClickEventType.Move)
-        dispatch({type: COMPUTE_DRAG, mouseButton: mouseButton === 1, point: point, shift: modifiers.shift || false})
-        _dispatchDragEvents(false)
+        dispatchDrag({type: COMPUTE_DRAG, mouseButton: mouseButton === 1, point: point, shift: modifiers.shift || false})
         _dispatchDiscreteMouseEvents(e, ClickEventType.Move)
-    }, [_dispatchDragEvents, _dispatchDiscreteMouseEvents])
+    }, [_dispatchDiscreteMouseEvents])
```
jsoules, line 137: Does the drag region still animate correctly after removing the `_dispatchDragEvents()` call here? (Like the draglayer in `ElectrodeGeometryTest`?) Or does the `useEffect()` hook pick it up?

jsoules (commented):
Couple questions I was unclear on.

PR #153, opened 2020-12-08
impl timeseriesview in jupyter

magland:


jupyterlab/labbox_ephys_widgets_jp/src/extensions/timeseries/TimeseriesViewNew/TimeseriesViewNew.tsx
```diff
@@ -71,7 +71,7 @@ const TimeseriesViewNew = (props: Props) => {
                     const slot = await timeseriesCalculationPool.requestSlot();
                     try {
                         result = await hither.createHitherJob(
-                            'get_timeseries_segment',
+                            'createjob_get_timeseries_segment',
```
magland, line 74: The jupyter extension only supports the new kind of hither calls where we have a createjob function that in turn creates and returns a hither job.

jupyterlab/labbox_ephys_widgets_jp/src/extensions/timeseries/timeseries.py
```diff
@@ -1,7 +1,12 @@
 import labbox_ephys as le
 
 
-@le.register_recording_view
-def Timeseries(*, sorting: le.LabboxEphysSortingExtractor, recording: le.LabboxEphysRecordingExtractor):
+@le.register_recording_view('Timeseries')
```
magland, line 4: register the timeseries widget as a recording view (accessible from python)

jupyterlab/labbox_ephys_widgets_jp/src/extensions/timeseries/timeseries.py
```diff
@@ -1,7 +1,12 @@
...
+def Timeseries_recordingview(*, recording: le.LabboxEphysRecordingExtractor):
+    import labbox_ephys_widgets_jp as lew
+    return lew.create_recording_view('TimeseriesView', recording=recording)
+
+@le.register_sorting_view('Timeseries')
```
magland, line 9: Also register it as a sorting view. In the future we will highlight spike times in the timeseries view.

jupyterlab/labbox_ephys_widgets_jp/src/pluginWidgets.tsx
```diff
@@ -12,21 +12,23 @@ import { sleepMsec } from './extensions/common/misc';
 import { activate as activatecorrelograms } from './extensions/correlograms/correlograms';
 import { activate as activateelectrodegeometry } from './extensions/electrodegeometry/electrodegeometry';
 import { ExtensionContext, HitherContext, HitherJob, HitherJobOpts, Recording, RecordingViewPlugin, Sorting, SortingUnitMetricPlugin, SortingUnitViewPlugin, SortingViewPlugin } from './extensions/extensionInterface';
+import { activate as activatetimeseries } from './extensions/timeseries/timeseries';
```
magland, line 15: activate the timeseries extension. In the future this section will be auto-generated from a template, similar to what is done for the web gui.

python/labbox_ephys/misc/calculate_timeseries_info.py
```diff
@@ -7,6 +7,17 @@
...
+@hi.function('createjob_calculate_timeseries_info', '0.1.2')
+def createjob_calculate_timeseries_info(labbox, recording_object):
+    jh = labbox.get_job_handler('timeseries')
+    jc = labbox.get_default_job_cache()
+    with hi.Config(
+        job_cache=jc,
+        job_handler=jh,
+        container=jh.is_remote
+    ):
+        return calculate_timeseries_info.run(recording_object=recording_object)
```
magland, line 19: Return a hither job. This is the new method of creating hither jobs from the javascript. Only the new method is supported in the jupyter extension.

src/extensions/timeseries/__init__.py
```diff
@@ -0,0 +1 @@
+from .timeseries import *
```
magland, line 1: This is just to ensure that all the widgets get registered via decorator. It doesn't actually export anything in particular for consumption.

src/extensions/timeseries/timeseries.py
```diff
@@ -1,7 +1,12 @@
 import labbox_ephys as le
 
 
-@le.register_recording_view
-def Timeseries(*, sorting: le.LabboxEphysSortingExtractor, recording: le.LabboxEphysRecordingExtractor):
+@le.register_recording_view('Timeseries')
```
magland, line 4: The optional parameter of register_recording_view is the name of the plugin (how it will be called from python). If omitted it would be equal to the function name. But in this case we have two different plugins with the same name, so we use this argument.

PR #157, opened 2020-12-09
implement in_memory option for le rec/sort extractors

magland:


devel/example_notebooks/example1.ipynb
```diff
@@ -23,43 +23,9 @@
...
-    "\n",
-    "        sorting_path = tmpdir + '/sorting.h5'\n",
-    "        le.extractors.H5SortingExtractorV1.write_sorting(sorting, sorting_path)\n",
-    "        S = le.LabboxEphysSortingExtractor({\n",
-    "            'sorting_format': 'h5_v1',\n",
-    "            'data': {\n",
-    "                'h5_path': kp.store_file(sorting_path)\n",
-    "            }\n",
-    "        })\n",
+    "    recording, sorting = se.example_datasets.toy_example(duration=50, num_channels=8, K=5, seed=1)\n",
```
magland, line 26: This example has been greatly simplified because we can now create a LabboxEphysRecordingExtractor/LabboxEphysSortingExtractor from non-le extractors.

devel/example_notebooks/example1.ipynb
```diff
@@ -23,43 +23,9 @@
...
+    "    R = le.LabboxEphysRecordingExtractor.from_memory(recording, serialize=True, serialize_dtype=np.int16)\n",
```
magland, line 27: We can specify whether or not to serialize the recording/sorting to disk. If so, the job cache will have an effect between different python sessions. Of course there is a disk space penalty.

magland: If serialization is not used, then the object is referenced in-memory, and the job cache only applies to the current process.

python/labbox_ephys/extractors/_in_memory.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def _random_string(num_chars: int) -> str:
+    chars = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789'
+    return ''.join(random.choice(chars) for _ in range(num_chars))
+
+unique_process_id = _random_string(20)
+in_memory_objects = {}
+
+def register_in_memory_object(obj):
```
magland, line 11: This is a way to register and retrieve in-memory objects. The dict() that is returned will be unique to that object for this python process.

python/labbox_ephys/extractors/labboxephysrecordingextractor.py
```diff
@@ -254,6 +258,11 @@ def __init__(self, arg: Union[str, dict], download: bool=False):
         elif recording_format == 'filtered':
             R = LabboxEphysRecordingExtractor(data['recording'], download=download)
             self._recording: se.RecordingExtractor = _apply_filters(recording=R, filters=data['filters'])
+        elif recording_format == 'in_memory':
```
magland, line 261: in_memory is now a special format for recordings/sortings

jsoules (approved):
This all looks fine to me. Checked out in Jupyter and web-served labbox, and no problems apparent.

PR #164, opened 2020-12-10
implement selection and selectionDispatch

magland:


devel/example_notebooks/example1.ipynb
```diff
@@ -80,7 +80,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "recording_views.Timeseries(recording=recording)"
+    "sorting_views.Curation(sorting=sorting, recording=recording)"
```
magland, line 83: Uncomment this line in the notebook

jupyterlab/labbox_ephys_widgets_jp/src/extensions/averagewaveforms/AverageWaveformsNew.tsx
```diff
@@ -20,11 +21,11 @@ const AverageWaveformsNew: React.FunctionComponent<SortingViewProps> = (props) =
     //     }
     // }, [hither, setResult, calcMode, setCalcMode])
     // return <div>AAAA: {calcMode} {result + ''}</div>
+    const selectedUnitIdsLookup: {[key: string]: boolean} = selection.selectedUnitIds.reduce((m, uid) => {m[uid + ''] = true; return m}, {} as {[key: string]: boolean})
```
magland, line 24: The old system was {[key: string]: boolean} which is fast and convenient for lookups, but not the best for storing a set of unit ids. The new system is selection.selectedUnitIds as a number[]. But here we need the lookup, so this code translates from new to old. Wondering if there is something more elegant than this .reduce() call.

jupyterlab/labbox_ephys_widgets_jp/src/extensions/averagewaveforms/AverageWaveformsNew.tsx
```diff
@@ -20,11 +21,11 @@ const AverageWaveformsNew: React.FunctionComponent<SortingViewProps> = (props) =
...
     return (
         <PlotGrid
             sorting={props.sorting}
-            selections={props.selectedUnitIds}
-            focus={props.focusedUnitId}
+            selections={selectedUnitIdsLookup}
```
magland, line 28: PlotGrid will still use the old system for now.

jupyterlab/labbox_ephys_widgets_jp/src/extensions/common/PlotGrid.tsx
```diff
@@ -21,7 +20,7 @@ interface Props {
     hither: HitherContext
 }
 
-const PlotGrid: FunctionComponent<Props> = ({ sorting, onUnitClicked, selections, focus,
+const PlotGrid: FunctionComponent<Props> = ({ sorting, onUnitClicked, selections,
```
magland, line 23: i am removing the focusedUnitId throughout for now, because it simplifies things... and if we find we need it, we can put it back in. Right now the only thing it was being used for (besides highlighting purposes) was to establish the anchor unit id for shift+click selection. But I now handle that internal to the table, so it doesn't need to be tracked outside of that component.

jupyterlab/labbox_ephys_widgets_jp/src/extensions/curation/CurationSortingView/ControlPanel.tsx
```diff
@@ -1,43 +1,44 @@
...
 import React, { FunctionComponent, useCallback } from 'react';
-import { SortingCuration, SortingCurationDispatch } from '../../extensionInterface';
+import { SortingCuration, SortingCurationDispatch, SortingSelection, SortingSelectionDispatch } from '../../extensionInterface';
 
 type Props = {
     width: number
     curation: SortingCuration
     curationDispatch: SortingCurationDispatch
-    selectedUnitIds: number[]
+    selection: SortingSelection
```
magland, line 9: replacing selectedUnitIds and onSelectedUnitIdsChanged by `selection` and `selectionDispatch` throughout

jupyterlab/labbox_ephys_widgets_jp/src/extensions/extensionInterface.ts
```diff
@@ -119,6 +119,48 @@ export const sortingCurationReducer: Reducer<SortingCuration, SortingCurationAct
 }
 ////////////////////
 
+// Sorting selection
```
magland, line 122: Implementation of new selection/selectionDispatch -- similar structure as curation/curationDispatch

jupyterlab/labbox_ephys_widgets_jp/src/extensions/extensionInterface.ts
```diff
@@ -150,11 +192,10 @@ export interface HitherContext {
 export interface SortingViewProps {
     sorting: Sorting
     recording: Recording
-    selectedUnitIds: {[key: string]: boolean}
-    focusedUnitId: number | null
     onUnitClicked: (unitId: number, event: {ctrlKey?: boolean, shiftKey?: boolean}) => void
     curationDispatch: (action: SortingCurationAction) => void
-    onSelectedUnitIdsChanged: (selectedUnitIds: {[key: string]: boolean}) => void
+    selection: SortingSelection
```
magland, line 196: simplified SortingView plugin props

jupyterlab/labbox_ephys_widgets_jp/src/extensions/timeseries/TimeWidgetNew/Splitter.tsx
```diff
@@ -25,9 +25,9 @@ const Splitter: FunctionComponent<Props> = (props) => {
     let child1 = props.children[0] as any as ReactElement
     let child2 = props.children[1] as any as ReactElement
 
-    const gripWidth = props.gripWidth !== undefined ? props.gripWidth : 12
-    const gripInnerWidth = props.gripInnerWidth !== undefined ? props.gripInnerWidth : 4
-    const gripMargin = props.gripMargin !== undefined ? props.gripMargin : 8
+    const gripWidth = props.gripWidth ?? 12
```
magland, line 28: thanks jeff for the ?? hint

jupyterlab/labbox_ephys_widgets_jp/src/extensions/unitstable/Units/Units.tsx
```diff
@@ -121,25 +121,22 @@ const Units: React.FunctionComponent<SortingViewProps & OwnProps> = (props) => {
...
-        Object.keys(selectedRowKeys).forEach((key) => selectedRowKeys[key]
-            ? labels.forEach((label) => handleAddLabel(Number(key), label))
-            : {});
-    };
-    const handlePurgeLabels = (selectedRowKeys: {[key: string]: any}, labels: Label[]) => {
-        Object.keys(selectedRowKeys).forEach((key) => selectedRowKeys[key]
-            ? labels.forEach((label) => handleRemoveLabel(Number(key), label))
-            : {});
-    };
+    // const handleAddLabel = (unitId: number, label: Label) => {
```
magland, line 124: No longer handling curation within the Units table

jupyterlab/labbox_ephys_widgets_jp/src/extensions/unitstable/Units/UnitsTable.tsx
```diff
@@ -109,12 +102,22 @@ const interpretSortFields = (fields: string[]): sortFieldEntry[] => {
...
 
 const UnitsTable: FunctionComponent<Props> = (props) => {
-    const { sortingUnitMetrics, units, metrics, selectedUnitIds, sorting, onSelectedUnitIdsChanged } = props
+    const { sortingUnitMetrics, units, metrics, selection, selectionDispatch, sorting } = props
     const sortingUnitMetricsList = sortByPriority(Object.values(sortingUnitMetrics)).filter(p => (!p.disabled))
     const [sortFieldOrder, setSortFieldOrder] = useState<string[]>([])
     units.sort((a, b) => a - b) // first sort by actual unit number
     // Now sort the list iteratively by each of the sorters in the sortFieldOrder state.
 
+    const toggleSelectedUnitId = useCallback(
```
magland, line 111: At some point, toggle can be handled in the sortingSelectionReducer for better performance. But for now this will be fine.

jupyterlab/labbox_ephys_widgets_jp/src/pluginWidgets.tsx
```diff
@@ -214,18 +214,18 @@ const PluginComponentWrapper: FunctionComponent<PluginComponentWrapperProps> = (
 
   const [curation, curationDispatch] = useReducer(sortingCurationReducer, defaultSortingCuration)
   sorting.curation = curation
+
+  const [selection, selectionDispatch] = useReducer(sortingSelectionReducer, defaultSortingSelection)
```
magland, line 218: Here's where we maintain the selection state for the jupyter widget. In the future we'll want to attach this to a python-defined context where we can share the selection state between different jupyter widgets.

src/containers/SortingView.tsx
```diff
@@ -47,16 +47,15 @@ type Props = StateProps & DispatchProps & OwnProps & RouteComponentProps
...
-
 const SortingView: React.FunctionComponent<Props> = (props) => {
   const { sortingViews, sortingUnitViews, documentInfo, sorting, sortingId, recording, curationDispatch, onSetSortingInfo, onSetExternalUnitMetrics, extensionsConfig, sortingUnitMetrics, hither } = props
   const { documentId, feedUri, readOnly } = documentInfo;
   const [sortingInfoStatus, setSortingInfoStatus] = useState<CalcStatus>('waiting');
   const [externalUnitMetricsStatus, setExternalUnitMetricsStatus] = useState<CalcStatus>('waiting');
   // const [selection, dispatchSelection] = useReducer(updateSelections, {focusedUnitId: null, selectedUnitIds: {}});
-  const [selectedUnitIds, setSelectedUnitIds] = useState<SelectedUnitIds>({})
-  const [focusedUnitId, setFocusedUnitId] = useState<number | null>(null)
+  const [selection, selectionDispatch] = useReducer(sortingSelectionReducer, defaultSortingSelection)
```
magland, line 56: Here is where we maintain the selection state for the sorting. This never gets to the global redux state, and I think that is okay for now.

src/extensions/averagewaveforms/AverageWaveformsNew.tsx
```diff
@@ -7,6 +7,7 @@ import AverageWaveformPlotNew from './AverageWaveformPlotNew';
 const averageWaveformsCalculationPool = createCalculationPool({maxSimultaneous: 6});
 
 const AverageWaveformsNew: React.FunctionComponent<SortingViewProps> = (props) => {
+    const { selection } = props
```
magland, line 10: These are repeat changes due to the code sync. I'm sort of wondering if we should avoid checking in both sets of files into the repo. I guess the sync could be part of the build step. I'm not sure. We'll leave it as is for now.

src/containers/SortingView.tsx
```diff
@@ -102,40 +101,31 @@ const SortingView: React.FunctionComponent<Props> = (props) => {
...
-              newSelectedUnitIds[uid] = true
-            }
+    else if ((event.shiftKey) && (anchorUnitId !== null)) {
+      const unitIds = intrange(anchorUnitId, unitId)
+      const newSelectedUnitIds = [...selection.selectedUnitIds]
+      const sortingInfo = sorting?.sortingInfo
+      if (sortingInfo) {
+        for (let uid of unitIds) {
+          if (sortingInfo.unit_ids.includes(uid)) {
+            if (!newSelectedUnitIds.includes(uid)) newSelectedUnitIds.push(uid)
```
jsoules, line 118: There's no reason to think we would have non-numeric or non-sorted unit IDs, right? I always assumed that the order they display on the screen would make sense relative to the sequence they appeared in the list, but I suppose this rule could give us problems if the units came in in some arbitrary unsorted order.

Of course we can probably just sort them, anyway. I just had a sudden concern about the edge case.

src/extensions/averagewaveforms/AverageWaveformsNew.tsx
jsoules: Yeah, maybe some kind of build hook would be a good idea? As long as unison is able to harmonize the files successfully. (I would hope that we won't have too much temptation to edit both copies of the file, but I guess you never know.)

jsoules (commented):
Looks good; not sure how much effort we want to put into the impedance mismatch between the way the selected units are represented (list vs dict). That and the debug line were probably my only comments, this otherwise looks good.

PR #167, opened 2020-12-11
develop curation view

magland:


.devcontainer/Dockerfile
```diff
@@ -100,6 +100,10 @@ RUN pip install matplotlib mpld3
 # nano
 RUN apt-get install -y nano
 
+# unison and unison-fsmonitor
```
magland: Not exactly sure why this change is in this PR, but i'm not going to fret.

jupyterlab/labbox_ephys_widgets_jp/src/extensions/curation/CurationSortingView/AllUnitsTab.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { FunctionComponent } from 'react';
```
magland, line 1: Implemented the AllUnits tab which shows the SortingView plugins. Maybe we want a better name than "All Units"

jupyterlab/labbox_ephys_widgets_jp/src/extensions/curation/CurationSortingView/SelectedUnitsTab.tsx
```diff
@@ -0,0 +1,55 @@
+import React, { Fragment, FunctionComponent } from 'react';
```
magland, line 1: Implemented the SlectedUnits tab which shows the SortingUnitView plugins for the selected units. Maybe we want a better name than "Selected Units"

jupyterlab/labbox_ephys_widgets_jp/src/extensions/curation/CurationSortingView/TabsView.tsx
```diff
@@ -0,0 +1,84 @@
+import AppBar from '@material-ui/core/AppBar';
```
magland, line 1: Implemented this TabsView which is the right-hand window in the curation view

jupyterlab/labbox_ephys_widgets_jp/src/extensions/devel/IndividualUnits/IndividualUnit.tsx
```diff
@@ -14,7 +14,7 @@ interface Props {
     width: number
     calculationPool: CalculationPool
     sortingInfo: SortingInfo
-    sortingUnitViews: {[key: string]: SortingUnitViewPlugin}
+    plugins: Plugins
```
magland, line 17: I simplified and renamed how registered plugins get passed around. Prior to the change, extensionContext meant two different things -- it referred to the object passed to the extensions that could be used to register plugins, and it also referred to the redux state that contained the registered plugins. I changed the latter to 'plugins' instead. And now 'plugins' state gets passed around as opposed to separately passing around SortingUnitViewPlugins, SortingViewPlugins, etc.

src/containers/SortingUnitView.tsx
```diff
@@ -78,7 +78,7 @@ const SortingUnitView: FunctionComponent<Props> = ({ sortingId, unitId, sorting,
             calculationPool={calculationPool}
             width={width}
             sortingInfo={sortingInfo}
-            sortingUnitViews={sortingUnitViews}
+            plugins={plugins}
```
magland, line 81: Most of these changes relate to the change described in my above comment... renaming extensionContext redux state to plugins redux state, and passing plugins object rather than sortingViewPlugins, etc.

jsoules (approved):
Assuming we can get the graphical weirdnesses in Jupyter worked out--or that we decide to just live with it until we can reimplement everything in Canvas--this should be good to go.

PR #171, opened 2020-12-11
extensions/ code sync

magland:


jupyterlab/labbox_ephys_widgets_jp/sync_extensions_code.py
```diff
@@ -0,0 +1,60 @@
+import os
+import shutil
+
+
+def sync_extensions_code():
```
magland, line 5: Here's the script that synchronizes the extensions/ folder

Checks if the folders are identical. If so we are good. If not:

If we are in a dev workspace (as evidenced by a file existing), then we raise an exception.
If we are not in a dev workspace, then we remove any existing extensions/ folder and replace it with a new one.

.gitignore
```diff
@@ -1,3 +1,5 @@
+jupyterlab/labbox_ephys_widgets_jp/src/extensions
```
magland, line 1: No we gitignore the jupyterlab/labbox_ephys_widgets_jp/src/extensions

and so we don't duplicate code in the repo!

.gitignore
```diff
@@ -1,3 +1,5 @@
...
+jupyterlab/codesync.txt
```
magland, line 2: We also ignore this file which is used to indicate whether we are in a development workspace

jupyterlab/labbox_ephys_widgets_jp/setup.py
```diff
@@ -50,6 +47,12 @@
     ('etc/jupyter/nbconfig/notebook.d' , HERE, 'labbox_ephys_widgets_jp.json')
 ]
 
+################################################################################
+# extensions code sync
+# Checks if the folders are identical. If so we are good. If not:
+# If we are in a dev workspace (as evidenced by a file existing), then we raise an exception.
+# If we are not in a dev workspace, then we remove any existing extensions/ folder and replace it with a new one.
+sync_extensions_code()
```
magland, line 55: Call this on `pip install` of the jupyter extension

jsoules (approved):
This all looks good. I gave it a test drive (by doing `pip install -e jupyterlab/labbox_ephys_widgets_jp`) for the following cases:

- dev mode, sync running --> Install runs successfully
- dev mode, sync not running, differences exist --> Install shows the expected error message 
- non-dev mode, differences exist --> Silently overwrites, then install succeeds
- non-dev mode, no differences --> Install succeeds

Although we should still look for a way to get pip to show us the output when install is successful (it'd be convenient to have the confirmation), this is the intended behavior so I'm going to go ahead and merge this.

PR #177, opened 2020-12-14
canvaswidget layers get their own props

magland:
In order to make layers that are shareable between multiple components (for example elec geom layer being behind multiple views), it is important that the prop types for layers are allowed to be different for different layers within the same canvas widget.

### DIFF
--- a/jupyterlab/labbox_ephys_widgets_jp/src/raindrop.tsx
+++ b/jupyterlab/labbox_ephys_widgets_jp/src/raindrop.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import CanvasWidget from './extensions/CanvasWidget';
 import { CanvasPainter } from './extensions/CanvasWidget/CanvasPainter';
-import { CanvasWidgetLayer, ClickEvent, ClickEventType, useCanvasWidgetLayer, useCanvasWidgetLayers } from './extensions/CanvasWidget/CanvasWidgetLayer';
+import { CanvasWidgetLayer, ClickEvent, ClickEventType } from './extensions/CanvasWidget/CanvasWidgetLayer';
 import { TransformationMatrix, Vec2 } from './extensions/CanvasWidget/Geometry';
 
 interface Raindrop {
@@ -82,8 +82,8 @@ interface Props {
 }
 
 const RaindropWidget: React.FunctionComponent<Props> = () => {
-  const layer = useCanvasWidgetLayer(createLayer)
-  const layers = useCanvasWidgetLayers([layer])
+  const layer = useLayer(createLayer)
+  const layers = useLayers([layer])
   return (
     <CanvasWidget
       layers={layers}
--- a/src/extensions/CanvasWidget/CanvasWidget.tsx
+++ b/src/extensions/CanvasWidget/CanvasWidget.tsx
@@ -1,5 +1,5 @@
 import React, { useCallback, useEffect, useReducer, useState } from 'react'
-import { BaseLayerProps, CanvasWidgetLayer, ClickEventType, formClickEventFromMouseEvent, KeyEventType } from './CanvasWidgetLayer'
+import { CanvasWidgetLayer, ClickEventType, formClickEventFromMouseEvent, KeyEventType } from './CanvasWidgetLayer'
 import { Vec2, Vec4 } from './Geometry'
 
 // This class serves three purposes:
@@ -77,14 +77,15 @@ const dragReducer = (state: DragState, action: DragAction): DragState => {
     throw Error('Unexpected: unreachable, but keeps ESLint happy')
 }
 
-interface Props<T extends BaseLayerProps> {
-    layers: CanvasWidgetLayer<T, any>[] | null, // the layers to paint (each corresponds to a canvas html element)
+interface Props {
+    layers: (CanvasWidgetLayer<any, any> | null)[], // the layers to paint (each corresponds to a canvas html element)
+    width: number,
+    height: number
     preventDefaultWheel?: boolean // whether to prevent default behavior of mouse wheel
-    layerProps: T // props sent to the layer
 }
 
-const CanvasWidget = <T extends BaseLayerProps>(props: Props<T>) => {
-    const { layers, layerProps, preventDefaultWheel } = props
+const CanvasWidget = (props: Props) => {
+    const { layers, preventDefaultWheel, width, height } = props
 
     // To learn about callback refs: https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node
     const [divElement, setDivElement] = useState<HTMLDivElement | null>(null)
@@ -100,10 +101,9 @@ const CanvasWidget = <T extends BaseLayerProps>(props: Props<T>) => {
         // or when the layers (prop) has changed (or if preventDefaultWheel has changed)
         // we set the canvas elements on the layers and schedule repaints
         if (!divElement) return
-        if (!layers) return
         layers.forEach((L, i) => {
             const canvasElement = divElement.children[i]
-            if (canvasElement) {
+            if ((canvasElement) && (L)) {
                 if (preventDefaultWheel) {
                     canvasElement.addEventListener("wheel", (event) => {
                         event.preventDefault()
@@ -123,21 +123,12 @@ const CanvasWidget = <T extends BaseLayerProps>(props: Props<T>) => {
         dragging: false
     })
 
-    const { width, height } = layerProps
-
-    // set the layer props on the layers
-    useEffect(() => {
-        if (!layers) return
-        layers.forEach(L => {
-            L.updateProps(layerProps)
-        })
-    }, [layerProps, layers])
-
     // schedule repaint when width or height change
     useEffect(() => {
-        if (!layers) return
         layers.forEach(L => {
-            L.scheduleRepaint()
+            if (L) {
+                L.scheduleRepaint()
+            }
         })
     }, [width, height, layers])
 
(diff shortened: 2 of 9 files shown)


jupyterlab/labbox_ephys_widgets_jp/src/raindrop.tsx
```diff
@@ -82,8 +82,8 @@ interface Props {
 }
 
 const RaindropWidget: React.FunctionComponent<Props> = () => {
-  const layer = useCanvasWidgetLayer(createLayer)
-  const layers = useCanvasWidgetLayers([layer])
+  const layer = useLayer(createLayer)
```
magland, line 85: rename useCanvasWidgetLayer to useCanvasLayer

src/extensions/CanvasWidget/CanvasWidget.tsx
```diff
@@ -77,14 +77,15 @@ const dragReducer = (state: DragState, action: DragAction): DragState => {
     throw Error('Unexpected: unreachable, but keeps ESLint happy')
 }
 
-interface Props<T extends BaseLayerProps> {
-    layers: CanvasWidgetLayer<T, any>[] | null, // the layers to paint (each corresponds to a canvas html element)
+interface Props {
+    layers: (CanvasWidgetLayer<any, any> | null)[], // the layers to paint (each corresponds to a canvas html element)
```
magland, line 81: The layers can have different props

src/extensions/CanvasWidget/CanvasWidget.tsx
```diff
@@ -77,14 +77,15 @@ const dragReducer = (state: DragState, action: DragAction): DragState => {
...
+    width: number,
+    height: number
     preventDefaultWheel?: boolean // whether to prevent default behavior of mouse wheel
-    layerProps: T // props sent to the layer
 }
 
-const CanvasWidget = <T extends BaseLayerProps>(props: Props<T>) => {
-    const { layers, layerProps, preventDefaultWheel } = props
+const CanvasWidget = (props: Props) => {
```
magland, line 87: CanvasWidget no longer has the layer props as a generic, because the layers can have different props.

src/extensions/CanvasWidget/CanvasWidget.tsx
```diff
@@ -100,10 +101,9 @@ const CanvasWidget = <T extends BaseLayerProps>(props: Props<T>) => {
         // or when the layers (prop) has changed (or if preventDefaultWheel has changed)
         // we set the canvas elements on the layers and schedule repaints
         if (!divElement) return
-        if (!layers) return
         layers.forEach((L, i) => {
             const canvasElement = divElement.children[i]
-            if (canvasElement) {
+            if ((canvasElement) && (L)) {
```
magland, line 106: The other change is that individual layers passed in to CanvasWidget can be temporarily null, where as the layers prop is always a list now and is not allowed to be null. Previously the layers could be null, but if not null, the individual layers would need to be null. So this is why we now need to check if L !== null throughout.

src/extensions/CanvasWidget/CanvasWidget.tsx
```diff
@@ -77,14 +77,15 @@ const dragReducer = (state: DragState, action: DragAction): DragState => {
     throw Error('Unexpected: unreachable, but keeps ESLint happy')
 }
 
-interface Props<T extends BaseLayerProps> {
-    layers: CanvasWidgetLayer<T, any>[] | null, // the layers to paint (each corresponds to a canvas html element)
+interface Props {
+    layers: (CanvasWidgetLayer<any, any> | null)[], // the layers to paint (each corresponds to a canvas html element)
+    width: number,
```
magland, line 82: But the canvas widget still needs to know the width/height... so those are passed to both the CanvasWidget props and to the individual layer props.

src/extensions/CanvasWidget/CanvasWidget.tsx
```diff
@@ -209,25 +202,37 @@ const CanvasWidget = <T extends BaseLayerProps>(props: Props<T>) => {
...
+                }
             }
         }
     }, [layers])
 
+    // make sure that the layers have width/height that matches this canvas widget
+    for (const L of layers) {
+        if (L) {
+            const {width: layerWidth, height: layerHeight} = L.getProps()
+            if ((layerWidth !== width) || (layerHeight !== height)) {
```
magland, line 226: We check that the width/height sent to the canvaswidget is the same as those passed to each layer.

src/extensions/CanvasWidget/CanvasWidgetLayer.ts
```diff
@@ -102,6 +102,7 @@ export const formKeyboardEvent = (type: KeyEventType, e: React.KeyboardEvent<HTM
 export class CanvasWidgetLayer<LayerProps extends BaseLayerProps, State extends object> {
     _onPaint: OnPaint<LayerProps, State>
     _onPropsChange: OnPropsChange<LayerProps>
+    _runningOnPropsChange = false
```
magland, line 105: We make sure that setProps is not called during onPropsChange()

src/extensions/CanvasWidget/CanvasWidgetLayer.ts
```diff
@@ -139,11 +140,18 @@ export class CanvasWidgetLayer<LayerProps extends BaseLayerProps, State extends
...
-        this._onPropsChange(this, p)
+    setProps(p: LayerProps) { // this should either be called inside the useLayer hook or in the render method of the parent to the CanvasWidget
+        if (this._runningOnPropsChange) {
+            throw Error('Calling setProps inside onPropsChange is not allowed.')
+        }
+        if ((this._props === null) || (!shallowEqual(this._props, p))) {
+            this._props = p
+            this._pixelWidth = p.width
+            this._pixelHeight = p.height
+            this._runningOnPropsChange = true
```
magland, line 151: We make sure that setProps is not called during onPropsChange()

src/extensions/CanvasWidget/CanvasWidgetLayer.ts
```diff
@@ -260,29 +268,65 @@ export class CanvasWidgetLayer<LayerProps extends BaseLayerProps, State extends
     }
 }
 
-export const useCanvasWidgetLayer = <LayerProps extends BaseLayerProps, LayerState extends Object>(createLayer: () => CanvasWidgetLayer<LayerProps, LayerState>): CanvasWidgetLayer<LayerProps, LayerState> | null => {
+export const useLayer = <LayerProps extends BaseLayerProps, LayerState extends Object>(createLayer: () => CanvasWidgetLayer<LayerProps, LayerState>, layerProps?: LayerProps): CanvasWidgetLayer<LayerProps, LayerState> | null => {
```
magland, line 271: The layers now get their props in the call to useLayer. Now useLayer is a hook that will return a persistent layer object, but the layer props are re-set with every new render call. However, the onPropsChange is only called on the layer when the props actually change (according to a shallow check for modifications)

src/extensions/CanvasWidget/CanvasWidgetLayer.ts
```diff
@@ -260,29 +268,65 @@ export class CanvasWidgetLayer<LayerProps extends BaseLayerProps, State extends
...
-            }
+const listsMatch = (list1: any[], list2: any[]) => {
+    if (list1.length !== list2.length) return false
+    for (let i = 0; i < list1.length; i++) {
+        if (list1[i] !== list2[i]) return false
+    }
+    return true
+}
+
+export const useLayers = (layers: (CanvasWidgetLayer<any, any> | null)[]) => {
```
magland, line 292: useLayers is now somewhat different than the previous useCanvasWidgetLayers. It basically returns a persistent list of layers that only changes if the individual entries change. This is important to avoid CanvasWidget thinking that the layers props has changed even when the entries of the list have remained the same.

src/extensions/timeseries/TimeWidgetNew/TimeWidgetNew.tsx
```diff
@@ -192,6 +191,13 @@ const timeReducer = (state: TimeState, action: TimeAction): TimeState => {
     }
 }
 
+const plotMargins = {
```
magland, line 194: this belongs outside the render function

src/extensions/timeseries/TimeWidgetNew/TimeWidgetNew.tsx
```diff
@@ -202,12 +208,12 @@ const TimeWidgetNew = (props: Props) => {
...
-    const timeAxisLayer = useCanvasWidgetLayer(createTimeAxisLayer)
-    const panelLabelLayer = useCanvasWidgetLayer(createPanelLabelLayer)
-    const cursorLayer = useCanvasWidgetLayer(createCursorLayer)
+    const mainLayer = useLayer(createMainLayer)
+    const timeAxisLayer = useLayer(createTimeAxisLayer)
+    const panelLabelLayer = useLayer(createPanelLabelLayer)
+    const cursorLayer = useLayer(createCursorLayer)
 
-    const layers = useCanvasWidgetLayers([mainLayer, timeAxisLayer, panelLabelLayer, cursorLayer])
+    const allLayers = useLayers([mainLayer, timeAxisLayer, panelLabelLayer, cursorLayer])
```
magland, line 216: This is where it was important to use useLayers so that the list object is persistent between calls when the individual layers have not changed.

jsoules (commented):
I'd like to talk this through, as I get the impression it's restructuring some of the division of responsibilities between the components in a way I don't understand.

PR #178, opened 2020-12-14
timeseriesview: select electrodes

magland:
User now must select which electrodes they want to view in the timeseries view

### DIFF
--- a/src/containers/TimeseriesForRecordingView.tsx
+++ b/src/containers/TimeseriesForRecordingView.tsx
@@ -37,9 +37,11 @@ const TimeseriesForRecordingView: FunctionComponent<Props> = ({ recordingId, rec
       <div>
         <TimeseriesViewNew
           recordingObject={recording.recordingObject}
+          recordingInfo={recording.recordingInfo}
           width={width}
           height={height}
           hither={hither}
+          opts={{channelSelectPanel: true}}
         />
       </div>
     </div>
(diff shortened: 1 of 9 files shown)


src/containers/TimeseriesForRecordingView.tsx
```diff
@@ -37,9 +37,11 @@ const TimeseriesForRecordingView: FunctionComponent<Props> = ({ recordingId, rec
       <div>
         <TimeseriesViewNew
           recordingObject={recording.recordingObject}
+          recordingInfo={recording.recordingInfo}
           width={width}
           height={height}
           hither={hither}
+          opts={{channelSelectPanel: true}}
```
magland, line 44: This option controls whether the elec geom widget appears in a left panel, allowing user to select:

image

src/extensions/timeseries/TimeseriesViewNew/TimeseriesWidgetNew.tsx
```diff
@@ -187,18 +188,22 @@ const TimeseriesWidgetNew = (props: Props) => {
     }, [yScaleDispatch])
 
     useEffect(() => {
-        if (timeseriesModel !== prevTimeseriesModel) {
+        if ((timeseriesModel !== prevTimeseriesModel) || (visibleChannelIds !== prevVisibleChannelIds)) {
```
magland, line 191: Need to refresh when the visible channel ids have changed.

src/extensions/timeseries/TimeseriesViewNew/TimeseriesWidgetNew.tsx
```diff
@@ -14,7 +14,7 @@ interface Props {
     y_scale_factor: number
     width: number
     height: number
-    leftPanels: any[]
+    visibleChannelIds?: number[] | null
```
magland, line 17: new optional prop called visibleChannelIds that controls which channels are visible

PR #181, opened 2020-12-15
implement spikeamplitudes timeview

magland:


src/containers/SortingView.tsx
```diff
@@ -33,6 +34,7 @@ interface StateProps {
 
 interface DispatchProps {
   onSetSortingInfo: (a: {sortingId: string, sortingInfo: SortingInfo}) => void
+  onSetRecordingInfo: (a: {recordingId: string, recordingInfo: RecordingInfo}) => void
```
magland, line 37: Need to obtain recording info via hither call

src/containers/SortingView.tsx
```diff
@@ -97,6 +99,32 @@ const SortingView: React.FunctionComponent<Props> = (props) => {
     }
   }, [onSetExternalUnitMetrics, setExternalUnitMetricsStatus, externalUnitMetricsStatus, sorting, sortingId, hither])
 
+  const [recordingInfoStatus, setRecordingInfoStatus]= useState<'waiting' | 'calculating' | 'finished' | 'error'>('waiting')
```
magland, line 102: Retrieve the recording info via hither if not already present on the recording.

src/extensions/CanvasWidget/CanvasPainter.ts
```diff
@@ -265,6 +265,24 @@ export class CanvasPainter {
         this._context2D.fillText(txt, config.x, config.y);
         this._context2D.restore()
     }
+    drawMarker(center: Vec2, opts: {radius: number, pen?: Pen, brush?: Brush}) {
```
magland, line 268: Implement drawMarker which is similar to drawEllipse+fillEllipse, except the radius is given in pixels rather than in the coordinate system of the painter. In the future we can pass a shape argument, like circle, square, triangle.

This is needed by plots that show markers.

src/extensions/averagewaveforms2/AverageWaveformsView/AverageWaveformWidget.tsx
```diff
@@ -55,9 +55,10 @@ const AverageWaveformWidget: FunctionComponent<Props> = (props) => {
     }
     const electrodesLayer = useLayer(createElectrodesLayer, layerProps)
     const waveformLayer = useLayer(createWaveformLayer, layerProps)
+    const layers = useLayers([electrodesLayer, waveformLayer])
```
magland, line 58: Not strictly necessary to use useLayers, but it helps reduce react renders

src/extensions/curation/curation.tsx
```diff
@@ -13,7 +13,7 @@ export function activate(context: ExtensionContext) {
         priority: 1000,
         component: CurationSortingView,
         props: {
-            height: 500
+            height: 1000
```
magland, line 16: increasing the hard-coded height. Will need to remedy. But curation.tsx will be probably be replaced by mountainview anyway.

src/extensions/spikeamplitudes/SpikeAmplitudesView/SelectUnitsWidget.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { FunctionComponent } from 'react';
```
magland, line 1: Allows user to select units, simply by displaying the table with no metrics rows.

src/extensions/spikeamplitudes/SpikeAmplitudesView/SpikeAmplitudesPanel.ts
```diff
@@ -0,0 +1,155 @@
+import { Brush, CanvasPainter, Font, Pen } from "../../CanvasWidget/CanvasPainter"
```
magland, line 1: A panel for the TimeView that displays spike amplitudes over time.

src/extensions/spikeamplitudes/SpikeAmplitudesView/SpikeAmplitudesPanel.ts
```diff
@@ -0,0 +1,155 @@
...
+            }
+            catch(err) {
+                this._calculationError = err
+            }
+            this._updateHandler && this._updateHandler()
+            this._calculationScheduled = false
+        }, 10)
+    }
+    async _doCalculation() {
+        const result = await this.args.hither.createHitherJob(
```
magland, line 110: Here's where we make a hither call to retrieve the actual spike amplitudes.

src/extensions/spikeamplitudes/SpikeAmplitudesView/SpikeAmplitudesPanel.ts
```diff
@@ -0,0 +1,155 @@
...
+                const ampMax = Math.max(...amplitudeRanges.map(r => r?.max || 0))
+                return {min: ampMin, max: ampMax}
+            }
+        }
+        return this.amplitudeRange()
+    }
+    setYRange(yrange: {min: number, max: number}) {
+        this._yrange = yrange
+    }
+    setPanelGroup(panels: SpikeAmplitudesPanel[]) {
```
magland, line 89: We need the panel group (list of all the panels for this view) so that we can determine a single consistent y-scale.

magland: See autoYRange above

src/extensions/spikeamplitudes/SpikeAmplitudesView/SpikeAmplitudesPanel.ts
```diff
@@ -0,0 +1,155 @@
...
+                }
+            }
+        ).wait()
+        this._timepoints = result.timepoints as number[]
+        this._amplitudes = result.amplitudes as number[]
+        this._maxAmplitude = Math.max(...this._amplitudes)
+    }
+}
+
+class CombinedPanel {
```
magland, line 132: The CombinedPanel is a panel that combines a collection of panels into one (so they overlap).

src/extensions/spikeamplitudes/SpikeAmplitudesView/SpikeAmplitudesTimeWidget.tsx
```diff
@@ -0,0 +1,55 @@
...
+type Props = {
+    recording: Recording
+    sorting: Sorting
+    unitIds: number[]
+    width: number
+    height: number,
+    hither: HitherContext
+}
+
+const SpikeAmplitudesTimeWidget: FunctionComponent<Props> = ({ recording, sorting, unitIds, width, height, hither }) => {
```
magland, line 15: Here's the new widget

src/extensions/spikeamplitudes/SpikeAmplitudesView/SpikeAmplitudesView.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { FunctionComponent } from 'react'
+import sizeMe, { SizeMeProps } from 'react-sizeme'
+import { SortingViewProps } from "../../extensionInterface"
+import Splitter from '../../timeseries/TimeWidgetNew/Splitter'
+import SelectUnitsWidget from './SelectUnitsWidget'
+import SpikeAmplitudesTimeWidget from './SpikeAmplitudesTimeWidget'
+
+const SpikeAmplitudesView: FunctionComponent<SortingViewProps & SizeMeProps> = ({recording, sorting, selection, selectionDispatch, size, hither}) => {
```
magland, line 8: Here's the new view that includes a unit select window (left) and the spike amplitudes widget (right)

src/extensions/spikeamplitudes/spikeamplitudes.py
```diff
@@ -0,0 +1,49 @@
+from typing import Dict
+
+import hither as hi
+import kachery as ka
+import numpy as np
+
+
+@hi.function('createjob_fetch_spike_amplitudes', '0.1.0')
+def createjob_fetch_spike_amplitudes(labbox, recording_object, sorting_object, unit_id):
```
magland, line 9: This is the hither call that retrieves the actual spike amplitudes

src/extensions/spikeamplitudes/spikeamplitudes.tsx
```diff
@@ -0,0 +1,16 @@
+// LABBOX-EXTENSION: spikeamplitudes
+// LABBOX-EXTENSION-TAGS: jupyter
+
+import { ExtensionContext } from '../extensionInterface';
+import SpikeAmplitudesView from './SpikeAmplitudesView/SpikeAmplitudesView';
+
+
+export function activate(context: ExtensionContext) {
+    context.registerSortingView({
```
magland, line 9: Register the new sorting view

src/extensions/timeseries/TimeWidgetNew/TimeWidgetNew.tsx
```diff
@@ -25,7 +25,7 @@ export type TimeWidgetAction = ActionItem | DividerItem
 
 interface Props {
     panels: TimeWidgetPanel[]
-    customActions: TimeWidgetAction[]
+    customActions?: TimeWidgetAction[]
```
magland, line 28: Making some props optional

src/registerExtensions.ts
```diff
@@ -29,6 +30,7 @@ const registerExtensions = (context: ExtensionContext) => {
     activateaveragewaveforms2(context)
     activatecorrelograms(context)
     activateexample(context)
+    activatespikeamplitudes(context)
```
magland, line 33: This comes from the code generation

src/extensions/spikeamplitudes/SpikeAmplitudesView/SpikeAmplitudesPanel.ts
```diff
@@ -0,0 +1,155 @@
...
+
+// const calculationPool = createCalculationPool({maxSimultaneous: 6})
+
+const colorList = [
+    'blue',
+    'green',
+    'red',
+    'orange',
+    'purple'
+]
```
jsoules, line 12: Might need a longer list, given the number of units. Also we should look into choosing accessible palettes--we'd like the tool to be useful including to color-blind or low-vision users. I just spent about half an hour looking for something off-the-shelf for this and haven't found it yet, so we should probably just open up a ticket for it and do it later...

(ref https://blog.graphiq.com/finding-the-right-color-palettes-for-data-visualizations-fcd4e707a283 https://medium.com/carbondesign/color-palettes-and-accessibility-features-for-data-visualization-7869f4874fca and many others)

src/extensions/spikeamplitudes/SpikeAmplitudesView/SpikeAmplitudesPanel.ts
```diff
@@ -0,0 +1,155 @@
...
+const colorForUnitId = (unitId: number) => {
+    while (unitId < 0) unitId += colorList.length
```
jsoules, line 14: We have negative unit ids??

src/extensions/spikeamplitudes/SpikeAmplitudesView/SpikeAmplitudesPanel.ts
```diff
@@ -0,0 +1,155 @@
...
+    constructor(private args: {recording: Recording, sorting: Sorting, unitId: number, hither: HitherContext}) {
+    }
+    setTimeRange(timeRange: {min: number, max: number}) {
+        this._timeRange = timeRange
+    }
+    paint(painter: CanvasPainter, completenessFactor: number) {
+        const timeRange = this._timeRange
+        if (!timeRange) return
+        const font: Font = {pixelSize: 12, family: 'Arial'}
+        const color = colorForUnitId(this.args.unitId)
```
jsoules, line 38: I'm a little torn on this, because while it's good to have consistent coloring for each unit id, it'd also be pretty easy to wind up with a lot of unit IDs that are rendered with the same color, even though others are available.

Something to worry about for the future.

src/extensions/spikeamplitudes/SpikeAmplitudesView/SpikeAmplitudesPanel.ts
jsoules: Gotcha. Nice.

jsoules (commented):
Couple future issues, a couple questions, and one or two suggestions for right now. In general looks to be in fine shape.

src/containers/SortingView.tsx
magland: I agree, I think we should uniformize this as CalculationStatus or something. With typescript I think we can get away with

type CalculationStatus = 'waiting' | 'calculating' | ...

using strings, because the ts compiler will force correct spelling, etc.

src/extensions/spikeamplitudes/SpikeAmplitudesView/SpikeAmplitudesPanel.ts
magland: yeah this is important. We need a good, consistent way to generate colors for units, and colors for channels.

src/extensions/spikeamplitudes/SpikeAmplitudesView/SpikeAmplitudesPanel.ts
magland: I've only seen them as non-negative, but I don't think we should exclude negatives.

PR #182, opened 2020-12-15
implement mountain-view style widget

magland:


.vscode/labbox-snippets.code-snippets
```diff
@@ -166,5 +166,14 @@
             "}"
         ],
         "description": ""
+    },
+    "usecallback": {
```
magland, line 170: Add a new snippet for useCallback

src/extensions/averagewaveforms/averagewaveforms.tsx
```diff
@@ -15,12 +15,14 @@ export function activate(context: ExtensionContext) {
     context.registerSortingView({
         name: 'AverageWaveformsNew',
         label: 'Average waveforms new',
-        component: AverageWaveformsNew
+        component: AverageWaveformsNew,
+        singleton: true
```
magland, line 19: We tag a sorting view plugin as singleton if it should only appear once in the MountainView (or similar) gui's

src/extensions/mountainview/MountainView/CurationControl.tsx
```diff
@@ -0,0 +1,114 @@
+import { Grid, Paper } from '@material-ui/core';
```
magland, line 1: I essentially copied this over from the curation widget.

src/extensions/mountainview/MountainView/MountainView.tsx
```diff
@@ -0,0 +1,112 @@
+import React, { FunctionComponent, useCallback, useReducer } from 'react'
```
magland, line 1: Here's the new MountainView plugin view.

src/extensions/mountainview/MountainView/ViewContainer.tsx
```diff
@@ -0,0 +1,53 @@
+import React, { FunctionComponent, useEffect, useState } from 'react';
```
magland, line 1: The ViewContainer is the right part of MountainView. It shows all the open views in a tab widget.

src/extensions/mountainview/MountainView/ViewContainerTabBar.tsx
```diff
@@ -0,0 +1,67 @@
+import { IconButton, Tab, Tabs } from '@material-ui/core';
```
magland, line 1: Here's the tab bar for the view container.

src/extensions/mountainview/MountainView/ViewContainerTabBar.tsx
```diff
@@ -0,0 +1,67 @@
...
+            variant="scrollable"
+        >
+            {views.map(v => (
+                viewContainerTab(v, onViewClosed)
+            ))}
+        </Tabs>
+    )
+}
+
+const viewContainerTab = (view: View, onClose: (view: View) => void) => {
```
magland, line 42: This is an individual tab. There is a reason that is just a function rather than a function component. It is because <Tabs> expects the direct children to be <Tab> elements.

src/extensions/mountainview/MountainView/ViewLauncher.tsx
```diff
@@ -0,0 +1,65 @@
+import React, { Fragment, FunctionComponent, useCallback } from 'react';
```
magland, line 1: This is the view launcher (part of the left window of MountainView) that provides buttons for launching the different types of views, one for each view plugin.

src/extensions/mountainview/MountainView/ViewWidget.tsx
```diff
@@ -0,0 +1,21 @@
+import React, { FunctionComponent } from 'react';
```
magland, line 1: This is a wrapper to an individual plugin view widget.

src/extensions/mountainview/mountainview.tsx
```diff
@@ -0,0 +1,15 @@
+// LABBOX-EXTENSION: mountainview
+// LABBOX-EXTENSION-TAGS: jupyter
+
+import { ExtensionContext } from '../extensionInterface';
+import MountainView from './MountainView/MountainView';
+
+export function activate(context: ExtensionContext) {
+    context.registerSortingView({
```
magland, line 8: Register the new sorting view

src/extensions/spikeamplitudes/spikeamplitudes.py
```diff
@@ -21,7 +21,7 @@ def createjob_fetch_spike_amplitudes(labbox, recording_object, sorting_object, u
             unit_id=unit_id
         )
 
-def _compute_peak_channel_index_from_average_form(average_waveform):
+def _compute_peak_channel_index_from_average_waveform(average_waveform):
```
magland, line 24: Reorganize the computation just a bit

src/extensions/unitstable/Units/Units.tsx
```diff
@@ -146,7 +148,7 @@ const Units: React.FunctionComponent<SortingViewProps & OwnProps> = (props) => {
     }
 
     return (
-        <div style={{'width': '100%'}}>
+        <div style={{width: props.width || size.width || 300}}>
```
magland, line 151: I was having some trouble getting the width to be right. Will need to figure out a cleaner way to do this.

src/registerExtensions.ts
```diff
@@ -13,6 +13,7 @@ import { activate as activatetimeseries } from './extensions/timeseries/timeseri
 import { activate as activateaveragewaveforms } from './extensions/averagewaveforms/averagewaveforms'
 import { activate as activatepythonsnippets } from './extensions/pythonsnippets/pythonsnippets'
 import { activate as activateunitstable } from './extensions/unitstable/unitstable'
+import { activate as activatemountainview } from './extensions/mountainview/mountainview'
```
magland, line 16: code generated

jsoules (commented):
Only real comment is that the name 'MountainView', while it is succinct, is not very transparent to users or future developers who aren't familiar with that tool/project. I would recommend something like 'TabbedVisualizationsView' or 'RepeatingTabbedView' or something.

magland:
> Only real comment is that the name 'MountainView', while it is succinct, is not very transparent to users or future developers who aren't familiar with that tool/project. I would recommend something like 'TabbedVisualizationsView' or 'RepeatingTabbedView' or something.

I think you are right that we need a more descriptive name for this. Maybe MVSortingView, where MV refers to the mountainview layout? But it's clear that it's a view of a sorting.

PR #184, opened 2020-12-15
fix toolbar click problem for timeseries view

magland:
Straightforward fix -- simplifed the render structure using divs with explicit positioning styles.


jsoules (approved):
Yeah, much simpler.

PR #191, opened 2020-12-15
Prevent scroll wheel from scrolling time series widget without focus

jsoules:
This PR introduces support for mouse entry and exit events in the CanvasWidget framework. It also makes the `mainLayer` of the `TimeSeries` widget series track whether it has focus (defined as having received a click since the last time the mouse left the window); if it loses focus, then it will no longer respond to mouse events.

Unfortunately, the TimeSeriesWidget series still captures mouse wheel events regardless of whether the layer has focus--this results in the window not scrolling on mouse events when the mouse is over the widget, even though the widget is blurred and we would like the window to scroll. Probably the easiest solution to this would be to allow the `CanvasWidget` object, rather than the individual layers, to track its focus status and `preventDefault` on mouse wheel events if the widget is focused. Unfortunately, React seems to be attaching such events in a way that Chrome and Firefox think they should be passive elements, which means they are not eligible to execute a `preventDefault`. This is to improve wheel and touch scrolling performance in certain circumstances; however, it's inconvenient for us. It probably also indicates that the event listener is being attached to the wrong element, since we don't want the whole page to ignore the event, just the `CanvasWidget`.

However, in default of seeing an actual fix for this issue in the near future, I've elected to push this code--which does contain an improvement, viz, random wheel events no longer scale the time series unintentionally--and defer the complete fix for a later date.

### DIFF
--- a/src/extensions/CanvasWidget/CanvasWidget.tsx
+++ b/src/extensions/CanvasWidget/CanvasWidget.tsx
@@ -98,13 +98,21 @@ const CanvasWidget = (props: Props) => {
     }, [])
 
     useEffect(() => {
+        if (divElement) {
+            (divElement as any)['_hasFocus'] = hasFocus
+        }
+    }, [hasFocus, divElement])
+
+    useEffect(() => {
         // this should be called only when the divElement has been first set (above)
         // or when the layers (prop) has changed (or if preventDefaultWheel has changed)
         // we set the canvas elements on the layers and schedule repaints
         if (!divElement) return
         if (!layers) return
         const stopScrollEvent = (e: Event) => {
-            e.preventDefault()
+            if ((divElement as any)['_hasFocus']) {
+                e.preventDefault()
+            }
         }
         layers.forEach((L, i) => {
             const canvasElement = divElement.children[i]
--- a/src/extensions/CanvasWidget/CanvasWidget.tsx
+++ b/src/extensions/CanvasWidget/CanvasWidget.tsx
@@ -89,6 +89,7 @@ const CanvasWidget = (props: Props) => {
 
     // To learn about callback refs: https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node
     const [divElement, setDivElement] = useState<HTMLDivElement | null>(null)
+    const [hasFocus, setHasFocus] = useState<boolean>(false)
     const divRef = React.useCallback((elmt: HTMLDivElement) => {
         // this should get called only once after the div has been written to the DOM
         // we set this div element so that it can be used below when we set the canvas
@@ -101,13 +102,15 @@ const CanvasWidget = (props: Props) => {
         // or when the layers (prop) has changed (or if preventDefaultWheel has changed)
         // we set the canvas elements on the layers and schedule repaints
         if (!divElement) return
+        if (!layers) return
+        const stopScrollEvent = (e: Event) => {
+            e.preventDefault()
+        }
         layers.forEach((L, i) => {
             const canvasElement = divElement.children[i]
             if ((canvasElement) && (L)) {
                 if (preventDefaultWheel) {
-                    canvasElement.addEventListener("wheel", (event) => {
-                        event.preventDefault()
-                    })
+                    canvasElement.addEventListener("wheel", stopScrollEvent)
                 }
                 L.resetCanvasElement(canvasElement)
                 L.scheduleRepaint()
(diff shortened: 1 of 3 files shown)


src/extensions/CanvasWidget/CanvasWidget.tsx
```diff
@@ -101,13 +102,15 @@ const CanvasWidget = (props: Props) => {
         // or when the layers (prop) has changed (or if preventDefaultWheel has changed)
         // we set the canvas elements on the layers and schedule repaints
         if (!divElement) return
+        if (!layers) return
+        const stopScrollEvent = (e: Event) => {
+            e.preventDefault()
+        }
```
jsoules, line 108: Factored this out to a specific function in order to facilitate removing the event listener below. While that did not solve the problem, I left it as a separate item in case we want to try that some more.

src/extensions/CanvasWidget/CanvasWidgetLayer.ts
```diff
@@ -54,21 +39,48 @@ export interface CanvasDragEvent {
...
+    Enter = 'ENTER',
+    Leave = 'LEAVE',
+    Out = 'OUT'
+}
+
+export interface WheelEvent {
+    deltaY: number
+}
+
+
```
jsoules, line 66: I alphabetized these.

src/extensions/CanvasWidget/CanvasWidgetLayer.ts
```diff
@@ -54,21 +39,48 @@ export interface CanvasDragEvent {
...
 // But I couldn't figure out how to make the inheritance work right, so I bagged it.
 // Reader, know that the layer in question ought to be a self-reference for the layer that owns the handler:
 // this allows the handler functions to modify the parent function state without having direct reference
 // to values outside their own scope.
 export type DiscreteMouseEventHandler = (event: ClickEvent, layer: CanvasWidgetLayer<any, any>) => void
 export type DragHandler = (layer: CanvasWidgetLayer<any, any>,  dragEvent: CanvasDragEvent) => void
-export type WheelEventHandler = (event: WheelEvent, layer: CanvasWidgetLayer<any, any>) => void
 export type KeyboardEventHandler = (event: KeyboardEvent, layer: CanvasWidgetLayer<any, any>) => boolean // return false to prevent default
+export type MousePresenceEventHandler = (event: MousePresenceEvent, layer: CanvasWidgetLayer<any, any>) => void
+export type WheelEventHandler = (event: WheelEvent, layer: CanvasWidgetLayer<any, any>) => void
```
jsoules, line 76: Alphabetized here also.

src/extensions/timeseries/TimeWidgetNew/mainLayer.ts
```diff
@@ -13,6 +13,7 @@ interface LayerState {
     inverseTransformations: TransformationMatrix[]
     anchorTimepoint: number | null
     dragging: boolean
+    captureWheel: boolean
```
jsoules, line 16: Tracks whether the `mainLayer` element has focus, and thus should scale with mouse wheel events, or doesn't, and thus should pass them through (but doesn't).

src/extensions/CanvasWidget/CanvasWidget.tsx
```diff
@@ -89,25 +89,36 @@ const CanvasWidget = (props: Props) => {
...
     const divRef = React.useCallback((elmt: HTMLDivElement) => {
         // this should get called only once after the div has been written to the DOM
         // we set this div element so that it can be used below when we set the canvas
         // elements to the layers
         setDivElement(elmt)
     }, [])
 
+    useEffect(() => {
+        if (divElement) {
+            (divElement as any)['_hasFocus'] = hasFocus
```
magland, line 102: The mouse wheel event does not have access to the state. However, it does have access to the divElement. Therefore we can attach something to the divElement that can be read on the mouse wheel event to determine whether we preventDefault. This could also be done using useRef(). The latter would probably be better as we would not need to hack the (divElement as any).

src/extensions/CanvasWidget/CanvasWidget.tsx
```diff
@@ -89,25 +89,36 @@ const CanvasWidget = (props: Props) => {
...
+    }, [hasFocus, divElement])
+
     useEffect(() => {
         // this should be called only when the divElement has been first set (above)
         // or when the layers (prop) has changed (or if preventDefaultWheel has changed)
         // we set the canvas elements on the layers and schedule repaints
         if (!divElement) return
+        if (!layers) return
+        const stopScrollEvent = (e: Event) => {
+            if ((divElement as any)['_hasFocus']) {
```
magland, line 113: Here is where we check the divElement for the _hasFocus field. As stated above we do not have access to the state in this listener callback.

magland:
Merging.

PR #195, opened 2020-12-16
plugin views: respect height prop when provided

magland:


src/extensions/electrodegeometry/electrodegeometry.tsx
```diff
@@ -13,7 +13,7 @@ const zipElectrodes = (locations: number[][], ids: number[]) => {
     })
 }
 
-const ElectrodeGeometryRecordingView: FunctionComponent<RecordingViewProps> = ({recording}) => {
+const ElectrodeGeometryRecordingView: FunctionComponent<RecordingViewProps> = ({recording, width, height}) => {
```
magland, line 16: Use width/height props if they are provided.

PR #205, opened 2020-12-17
Implement max electrode radius of roughly 32 pixels.

jsoules:
Not super elegant--we wind up doing a rough guess of the scaling factor and then passing that in to the radius computation, so some of the work happens twice. Current max pixel radius hard-coded at 32 pixels; feel free to adjust as appropriate (it looks good on my screen, but your mileage may vary).

### DIFF
--- a/src/extensions/CanvasWidget/CanvasPainter.ts
+++ b/src/extensions/CanvasWidget/CanvasPainter.ts
@@ -162,6 +162,14 @@ export class CanvasPainter {
     exportingFigure() {
         return this._exportingFigure
     }
+    // This is basically just for debugging purposes.
+    fillWholeCanvas(color: Color): void {
+        console.log(`Pixel dimensions: 0 to ${this._pixelWidth} and 0 to ${this._pixelHeight}`)
+        this._context2D.save()
+        applyBrush(this._context2D, {color: color})
+        this._context2D.fillRect(0, 0, this._pixelWidth, this._pixelHeight)
+        this._context2D.restore()
+    }
     clear(): void {
         this.clearRect({xmin: 0, xmax: this._pixelWidth, ymin: 0, ymax: this._pixelHeight});
     }
--- a/src/extensions/electrodegeometry/ElectrodeGeometryWidget/ElectrodeGeometryWidget.tsx
+++ b/src/extensions/electrodegeometry/ElectrodeGeometryWidget/ElectrodeGeometryWidget.tsx
@@ -81,9 +81,7 @@ const initialElectrodeLayerState: ElectrodeLayerState = {
 }
 
 const computeRadiusCache = new Map<string, number>()
-const computeRadius = (electrodes: Electrode[], scalingFactor: number): number => {
-    const MAX_RADIUS_PIXELS = 32
-    const MAX_RADIUS_COORDS = MAX_RADIUS_PIXELS / scalingFactor
+const computeRadius = (electrodes: Electrode[]): number => {
     const key = JSON.stringify(electrodes)
     const val = computeRadiusCache.get(key)
     if (val !== undefined) {
@@ -102,8 +100,8 @@ const computeRadius = (electrodes: Electrode[], scalingFactor: number): number =
     })
     // (might set a hard cap, but remember these numbers are in electrode-space coordinates)
     const radius = 0.4 * leastNorm
-    computeRadiusCache.set(key, Math.min(radius, MAX_RADIUS_COORDS))
-    return Math.min(radius, MAX_RADIUS_COORDS)
+    computeRadiusCache.set(key, radius)
+    return radius
 }
 
 const getElectrodesBoundingBox = (electrodes: Electrode[], radius: number): RectangularRegion => {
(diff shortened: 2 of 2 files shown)


src/extensions/electrodegeometry/ElectrodeGeometryWidget/ElectrodeGeometryWidget.tsx
```diff
@@ -123,12 +125,23 @@ const onUpdateLayerProps = (layer: CanvasWidgetLayer<ElectrodeLayerProps, Electr
...
+    // close enough for our purposes.
+    // Get the electrode bounding box (assuming 0 radius), orient per the canvas aspect ratio orientation,
+    // then set the scaling factor per whether we are constrained by width or height. These steps will be repeated below, alas.
+    const rawElectrodeBoundingBox = getElectrodesBoundingBox(electrodes, 0)
+    let rawAspect = getWidth(rawElectrodeBoundingBox) / getHeight(rawElectrodeBoundingBox)
+    const aspectMismatch = ((rawAspect > 1) !== (canvasAspect > 1))
+    rawAspect = aspectMismatch ? 1/rawAspect : rawAspect
+    const roughScalingFactor = rawAspect > canvasAspect ? W / getWidth(rawElectrodeBoundingBox) : H / getHeight(rawElectrodeBoundingBox)
+
+    const radius = computeRadius(electrodes, roughScalingFactor)
```
magland, line 139: If (radius > ....) then adjust scale factor

magland: Ooops, not here, we need to use the pixelRadius for determining that.

magland:
If pixel radius exceeds a max, then we need to adjust the entire layout -- in other words, choose a scale factor such that the pixel radius does not exceed some maximum.

jsoules:
Okay, I've made some changes to this so that now the entire display scales down if the radius is over a certain size. I've confirmed that the result remains centered. I also added a method to `CanvasPainter` to assist in visualizing the canvas (to confirm that the elements I was displaying remain centered) and have a couple lines of test code to make it more clear what's happening under the hood for this widget.

Ready for re-review. (I think 32 px is a good radius size on my monitor at least, but if you'd like to see a dramatic effect, might bump it down to 16 so it's more obvious.)

src/extensions/CanvasWidget/CanvasPainter.ts
```diff
@@ -162,6 +162,14 @@ export class CanvasPainter {
...
         return this._exportingFigure
     }
+    // This is basically just for debugging purposes.
+    fillWholeCanvas(color: Color): void {
+        console.log(`Pixel dimensions: 0 to ${this._pixelWidth} and 0 to ${this._pixelHeight}`)
+        this._context2D.save()
+        applyBrush(this._context2D, {color: color})
+        this._context2D.fillRect(0, 0, this._pixelWidth, this._pixelHeight)
+        this._context2D.restore()
+    }
```
jsoules, line 172: This method added to `CanvasPainter` to facilitate visualizing the entire drawing field, in case we're wondering about layout issues.

src/extensions/electrodegeometry/ElectrodeGeometryWidget/ElectrodeGeometryWidget.tsx
```diff
@@ -150,6 +152,11 @@ const onUpdateLayerProps = (layer: CanvasWidgetLayer<ElectrodeLayerProps, Electr
         scaleFactor = H / getHeight(boundingBox)
     }
 
+    // We don't want to have big huge electrode circles if there are too few electrodes relative to the canvas.
+    // To correct for this, adjust the scale factor downwards if it would result in an unacceptably large radius.
+    const MAX_RADIUS_PIXELS = 32
```
jsoules, line 157: This constant could be defined elsewhere (at the least, higher up in the function) but for now I wanted to keep it close to where it was used so its function was more obvious.

src/extensions/electrodegeometry/ElectrodeGeometryWidget/ElectrodeGeometryWidget.tsx
```diff
@@ -175,6 +182,10 @@ const onUpdateLayerProps = (layer: CanvasWidgetLayer<ElectrodeLayerProps, Electr
 const paintElectrodeGeometryLayer = (painter: CanvasPainter, props: ElectrodeLayerProps, state: ElectrodeLayerState) => {
     painter.wipe()
     const useLabels = state.pixelRadius > 5
+    // The following three lines are visualizations to confirm the scaled image remains centered.
+    // painter.fillWholeCanvas('rgb(224, 196, 224)')
+    // const electrodesBoundngBox = getElectrodesBoundingBox(props.electrodes, 0)
+    // painter.drawRect(electrodesBoundngBox, {color: 'black', width: 4})
```
jsoules, line 188: These lines collectively paint the whole canvas a light pink and then draw a black box around the centers of the electrodes (exclusive of the radii). Just for visualization purposes.

magland:
Tested on my end - looks good. Merging.

PR #206, opened 2020-12-17
jupyter: export curation from MVSortingView object

### LINKED ISSUE #202: jupyter: export curation from MVSortingView object
For example:

X = le.sorting_views.MVSortingView(sorting=sorting, recording=recording)
display(X)

X.get_curation(...)

magland:
Fixes #202

### DIFF
--- a/jupyterlab/labbox_ephys_widgets_jp/labbox_ephys_widgets_jp/create_sorting_view.py
+++ b/jupyterlab/labbox_ephys_widgets_jp/labbox_ephys_widgets_jp/create_sorting_view.py
@@ -37,6 +37,8 @@ def create_sorting_view(plugin_name: str, *, sorting: le.LabboxEphysSortingExtra
         recordingObject = DictTrait(recording.object()).tag(sync=True)
         recordingInfo = DictTrait(le.get_recording_info(recording_object=recording.object())).tag(sync=True)
         sortingInfo = DictTrait(le.get_sorting_info(sorting_object=sorting.object(), recording_object=recording.object())).tag(sync=True)
+        selection = DictTrait({}).tag(sync=True)
+        curation = DictTrait({}).tag(sync=True)
         def __init__(self) -> None:
             super().__init__()
             self._jobs_by_id: Dict[str, HitherJob] = {}
--- a/jupyterlab/labbox_ephys_widgets_jp/src/pluginWidgets.tsx
+++ b/jupyterlab/labbox_ephys_widgets_jp/src/pluginWidgets.tsx
@@ -2,8 +2,8 @@
 // Distributed under the terms of the Modified BSD License.
 
 // Import the CSS
-import { DOMWidgetModel, DOMWidgetView, ISerializers } from '@jupyter-widgets/base';
-import React, { FunctionComponent, useReducer } from 'react';
+import { DOMWidgetModel, DOMWidgetView, ISerializers, WidgetModel } from '@jupyter-widgets/base';
+import React, { FunctionComponent, useEffect, useReducer } from 'react';
 import ReactDOM from 'react-dom';
 import '../css/widget.css';
 import exampleSorting from './exampleSorting';
@@ -161,7 +161,9 @@ export class SortingViewModel extends DOMWidgetModel {
       _view_module_version: SortingViewModel.view_module_version,
       pluginName: '',
       sortingObject: {},
-      recordingObject: {}
+      recordingObject: {},
+      curation: defaultSortingCuration,
+      selection: defaultSortingSelection
     };
   }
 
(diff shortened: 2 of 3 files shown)


jupyterlab/labbox_ephys_widgets_jp/labbox_ephys_widgets_jp/create_sorting_view.py
```diff
@@ -37,6 +37,8 @@ class SortingView(DOMWidget):
         recordingObject = DictTrait(recording.object()).tag(sync=True)
         recordingInfo = DictTrait(le.get_recording_info(recording_object=recording.object())).tag(sync=True)
         sortingInfo = DictTrait(le.get_sorting_info(sorting_object=sorting.object(), recording_object=recording.object())).tag(sync=True)
+        selection = DictTrait({}).tag(sync=True)
```
magland, line 40: Add new selection and curation traits to the jupyter python object. This will allow:

X = sorting_views.MVSortingView(...)
X.curation = set curation here

Or retrieve the curation in a different cell:
X.curation

jupyterlab/labbox_ephys_widgets_jp/src/pluginWidgets.tsx
```diff
@@ -161,7 +161,9 @@ export class SortingViewModel extends DOMWidgetModel {
       _view_module_version: SortingViewModel.view_module_version,
       pluginName: '',
       sortingObject: {},
-      recordingObject: {}
+      recordingObject: {},
+      curation: defaultSortingCuration,
```
magland, line 165: Attach these traits to the model on the typescript side

jupyterlab/labbox_ephys_widgets_jp/src/pluginWidgets.tsx
```diff
@@ -217,11 +220,39 @@ const PluginComponentWrapper: FunctionComponent<PluginComponentWrapperProps> = (
     recording = example.recording
   }
 
-  const [curation, curationDispatch] = useReducer(sortingCurationReducer, defaultSortingCuration)
-  sorting.curation = curation
+  // const [curation, curationDispatch] = useReducer(sortingCurationReducer, defaultSortingCuration)
 
-  const [selection, selectionDispatch] = useReducer(sortingSelectionReducer, defaultSortingSelection)
+  // curation
+  const [curation, curationDispatch] = useReducer(sortingCurationReducer, model.get('curation').labelChoices ? model.get('curation') : defaultSortingCuration)
```
magland, line 226: This code block is used to synchronize the curation between the jupyter system and the react state system.

jupyterlab/labbox_ephys_widgets_jp/src/pluginWidgets.tsx
```diff
@@ -217,11 +220,39 @@ const PluginComponentWrapper: FunctionComponent<PluginComponentWrapperProps> = (
...
+  }, [curation, model])
+  model.on('change:curation', () => {
+    curationDispatch({
+      type: 'SetCuration',
+      curation: model.get('curation')
+    })
+  }, null)
+
+  // selection
+  const [selection, selectionDispatch] = useReducer(sortingSelectionReducer, model.get('selection').selectedUnitIds ? model.get('selection') : defaultSortingSelection)
```
magland, line 241: Similar for the curation

jupyterlab/labbox_ephys_widgets_jp/src/pluginWidgets.tsx
```diff
@@ -277,6 +309,7 @@ export class SortingView extends DOMWidgetView {
         recordingInfo={recordingInfo}
         plugins={plugins}
         calculationPool={calculationPool}
+        model={this.model}
```
magland, line 312: Pass the jupyter widget model to the react component

src/extensions/extensionInterface.ts
```diff
@@ -72,6 +72,11 @@ export const defaultSortingCuration: SortingCuration = {
 
 export type SortingCurationDispatch = (action: SortingCurationAction) => void
 
+type SetCurationSortingCurationAction = {
```
magland, line 75: New action to set the curation all at once (needed to sync with jupyter widget)

jupyterlab/labbox_ephys_widgets_jp/src/pluginWidgets.tsx
jsoules: From Slack discussion, this is creating two hooks. One is the React `useEffect` hook which fires on changes to the curation object (or if the actual `model` reference were to change, which won't happen). This hook checks if the `model` object (a feature of the underlying Jupyter labextension framework) differs from the UI selection state. If there is a difference, the hook updates the `model` state to match the current UI state.

The second hook is attached to the `model` itself. On a change to its `curation` property, it calls the `curationDispatch` reducer with a new verb, `SetCuration`, which simply overwrites the existing curation value in the React state with whatever is in the `model`.

It is anticipated that changes would not be initiated in the `model` object, so effectively this just does a full circle to make sure everything is in sync any time a UI element changes. The conditional in line 228 (`if (model.get('curation') !== curation)`) ensures that there won't be any cyclic updates.

jsoules (approved):

jupyterlab/labbox_ephys_widgets_jp/src/pluginWidgets.tsx
magland: yes that's right

PR #208, opened 2020-12-17
implement merge curation

magland:


jupyterlab/labbox_ephys_widgets_jp/src/pluginWidgets.tsx
```diff
@@ -162,8 +162,8 @@ export class SortingViewModel extends DOMWidgetModel {
       pluginName: '',
       sortingObject: {},
       recordingObject: {},
-      curation: defaultSortingCuration,
-      selection: defaultSortingSelection
+      curation: {},
```
magland, line 165: No longer maintain a defaultSortingCuration and a defaultSortingSelection. Just use empty object for default, and all fields are optional.

src/actions/index.ts
```diff
@@ -42,8 +42,11 @@ export const SET_DOCUMENT_INFO = 'SET_DOCUMENT_INFO'
 
 export const SET_SERVER_INFO = 'SET_SERVER_INFO'
 
+export const SET_CURATION = 'SET_CURATION'
```
magland, line 45: 3 new curation actions SET_CURATION, MERGE_UNITS, UNMERGE_UNITS

src/containers/SortingView.tsx
```diff
@@ -130,16 +130,16 @@ const SortingView: React.FunctionComponent<Props> = (props) => {
 
   const handleUnitClicked = useCallback((unitId, event) => {
     if (event.ctrlKey) {
-      if (selection.selectedUnitIds.includes(unitId)) {
-        selectionDispatch({type: 'SetSelectedUnitIds', selectedUnitIds: selection.selectedUnitIds.filter(uid => (uid !== unitId))})
+      if ((selection.selectedUnitIds || []).includes(unitId)) {
```
magland, line 133: Since the fields on selection are all optional, we need to do a lot of '|| []' throughout. This could be avoided by making a `const selectedUnitIds = selection.selectedUnitIds || []` at the top of these functions. Will do that sometime in the future.

src/extensions/extensionInterface.ts
```diff
@@ -62,12 +62,9 @@ export interface MetricPlugin extends LabboxPlugin {
 
 // Sorting curation
 export type SortingCuration = {
-    labelsByUnit: {[key: string]: string[]}
-    labelChoices: string[]
-}
-export const defaultSortingCuration: SortingCuration = {
-    labelChoices: ["accept", "reject", "mua", "artifact", "noise"].sort(),
-    labelsByUnit: {}
+    labelsByUnit?: {[key: string]: string[]}
```
magland, line 65: These fields are now optional

src/extensions/extensionInterface.ts
```diff
@@ -62,12 +62,9 @@ export interface MetricPlugin extends LabboxPlugin {
...
+    labelChoices?: string[]
+    mergeGroups?: (number[])[]
```
magland, line 67: This is the new mergeGroups field of the curation. An example: `[[1, 2, 4], [3, 9]]`

src/extensions/extensionInterface.ts
```diff
@@ -88,8 +85,43 @@ type RemoveLabelSortingCurationAction = {
     unitId: number
     label: string
 }
-
-export type SortingCurationAction = SetCurationSortingCurationAction | AddLabelSortingCurationAction | RemoveLabelSortingCurationAction
+type MergeUnitsSortingCurationAction = {
```
magland, line 88: New action for merging units.

src/extensions/extensionInterface.ts
```diff
@@ -88,8 +85,43 @@ type RemoveLabelSortingCurationAction = {
...
+    type: 'MergeUnits',
+    unitIds: number[]
+}
+type UnmergeUnitsSortingCurationAction = {
```
magland, line 92: New action for unmerging units.

src/extensions/extensionInterface.ts
```diff
@@ -88,8 +85,43 @@ type RemoveLabelSortingCurationAction = {
...
+    type: 'UnmergeUnits',
+    unitIds: number[]
+}
+
+export type SortingCurationAction = SetCurationSortingCurationAction | AddLabelSortingCurationAction | RemoveLabelSortingCurationAction | MergeUnitsSortingCurationAction | UnmergeUnitsSortingCurationAction
+
+const intersection = (a: number[], b: number[]) => (
```
magland, line 99: intersection and union are set operations on lists

src/extensions/extensionInterface.ts
```diff
@@ -88,8 +85,43 @@ type RemoveLabelSortingCurationAction = {
...
+    a.filter(x => (b.includes(x)))
+)
+const union = (a: number[], b: number[]) => (
+    [...a, ...b.filter(x => (!a.includes(x)))].sort()
+)
+
+const simplifyMergeGroups = (mg: (number[])[]): (number[])[] => {
```
magland, line 106: simplifyMergeGroups is where the tricky work takes place. Some examples:

```
[[1, 2], [2, 3], [5, 6]] => [[1, 2, 3], [5, 6]]
[[1, 5], [3, 4], [3, 5]] => [[1, 3, 4, 5]]
[[5]] => []
```

src/extensions/extensionInterface.ts
```diff
@@ -124,18 +156,26 @@ export const sortingCurationReducer: Reducer<SortingCuration, SortingCurationAct
         }
         else return state
     }
+    else if (action.type === 'MergeUnits') {
+        return {
+            ...state,
+            mergeGroups: simplifyMergeGroups([...(state.mergeGroups || []), action.unitIds])
```
magland, line 162: To merge units, we just need to add a new merge group and then simplify

src/extensions/extensionInterface.ts
```diff
@@ -124,18 +156,26 @@ export const sortingCurationReducer: Reducer<SortingCuration, SortingCurationAct
...
+        }
+    }
+    else if (action.type === 'UnmergeUnits') {
+        return {
+            ...state,
+            mergeGroups: simplifyMergeGroups((state.mergeGroups || []).map(g => (g.filter(x => (!action.unitIds.includes(x))))))
```
magland, line 168: To unmerge units, we just need to remove the units from everywhere they appear in the merge groups, and then simplify.

src/extensions/extensionInterface.ts
```diff
@@ -124,18 +156,26 @@ export const sortingCurationReducer: Reducer<SortingCuration, SortingCurationAct
...
 
 // Sorting selection
 export type SortingSelection = {
-    selectedUnitIds: number[]
-    visibleUnitIds: number[] | null // null means all are selected
-}
-export const defaultSortingSelection: SortingSelection = {
-    selectedUnitIds: [],
-    visibleUnitIds: null
+    selectedUnitIds?: number[]
```
magland, line 177: These fields are now optional

src/extensions/mountainview/MVSortingView/CurationControl.tsx
```diff
@@ -59,18 +74,28 @@ const CurationControl: FunctionComponent<Props & SizeMeProps> = ({ selection, se
...
-        partial: labelCounts[label] < selection.selectedUnitIds.length ? true : false
+        partial: labelCounts[label] < selectedUnitIds.length ? true : false
     }))
     const paperStyle: React.CSSProperties = {
         marginTop: 25,
         marginBottom: 25,
         backgroundColor: '#f9f9ff'
     }
-    const enableApply = selection.selectedUnitIds.length > 0
+    const unitsAreInMergeGroups = (unitIds: number[]) => {
```
magland, line 84: Returns true if all the units are in some merge groups. We only want to give the "unmerge" option if this is the case.

src/extensions/mountainview/MVSortingView/CurationControl.tsx
```diff
@@ -59,18 +74,28 @@ const CurationControl: FunctionComponent<Props & SizeMeProps> = ({ selection, se
...
+        const mg = curation.mergeGroups || []
+        const all = mg.reduce((prev, g) => [...prev, ...g], []) // all units in merge groups
+        for (let unitId of unitIds) {
+            if (!all.includes(unitId)) return false
+        }
+        return true
+    }
+    const enableApply = selectedUnitIds.length > 0
+    const standardChoices = ['accept', 'reject', 'noise', 'artifact', 'mua']
```
magland, line 93: Always offer some standard choices for labels, and then also offer custom choices for labels.

src/extensions/mountainview/MVSortingView/CurationControl.tsx
```diff
@@ -84,22 +109,29 @@ const CurationControl: FunctionComponent<Props & SizeMeProps> = ({ selection, se
...
+                                </Grid>
+                            ): <span />
+                        )
+                    }
+                </Grid>
+            </Paper>
+            <Paper style={paperStyle} key="merge">                
+                Merge:
+                {
+                    selectedUnitIds.length >= 2 && <button key="merge" onClick={handleMergeSelected}>Merge selected units: {selectedUnitIds.join(', ')}</button>
```
magland, line 129: New merge button that appears when 2 or more units are selected.

src/extensions/mountainview/MVSortingView/CurationControl.tsx
```diff
@@ -84,22 +109,29 @@ const CurationControl: FunctionComponent<Props & SizeMeProps> = ({ selection, se
...
+                }
+                {
+                    (selectedUnitIds.length > 0 && unitsAreInMergeGroups(selectedUnitIds)) && <button key="unmerge" onClick={handleUnmergeSelected}>Unmerge units: {selectedUnitIds.join(', ')}</button>
```
magland, line 132: new unmerge button that appears if the selected units are all in merge groups.

src/extensions/unitstable/Units/UnitsTable.tsx
```diff
@@ -48,14 +48,14 @@ const UnitCheckbox = React.memo((a: {unitKey: string, selected: boolean, handleC
...
+const UnitIdCell = React.memo((a: {id: number, mergeGroup: number[] | null, sortingId: string}) => {
     // const elmt = (
     //     <Link to={`/${a.documentId}/sortingUnit/${a.sortingId}/${a.id}/${getPathQuery({feedUri: a.feedUri})}`}>
     //         {a.id}
     //     </Link>
     // )
-    const elmt = a.id
-    return <TableCell><span>{elmt}</span></TableCell>
+    const g = a.mergeGroup
+    return <TableCell><span>{a.id + ''}{g && ' (' + g.filter(x => (x !== a.id)).join(', ') + ')'}</span></TableCell>
```
magland, line 58: Display the merge group information next to the unit ID in the units table. Good luck parsing this line. Better to just see what it does in practice: 

image

src/extensions/unitstable/Units/UnitsTable.tsx
```diff
@@ -130,7 +130,11 @@ const UnitsTable: FunctionComponent<Props> = (props) => {
             return sortMetricValues(recordA, recordB, r.sortAscending)
         })
     }
-    const selectedUnitIdsLookup: {[key: string]: boolean} = selection.selectedUnitIds.reduce((m, uid) => {m[uid + ''] = true; return m}, {} as {[key: string]: boolean})
+    const selectedUnitIdsLookup: {[key: string]: boolean} = (selection.selectedUnitIds || []).reduce((m, uid) => {m[uid + ''] = true; return m}, {} as {[key: string]: boolean})
+    const mergeGroupForUnitId = (unitId: number) => {
```
magland, line 134: Get the merge group for the unit ID, or null.

src/extensions/extensionInterface.ts
```diff
@@ -88,8 +85,43 @@ type RemoveLabelSortingCurationAction = {
...
+            for (let j = i + 1; j < newMergeGroups.length; j ++) {
+                const g2 = newMergeGroups[j]
+                if (intersection(g1, g2).length > 0) {
+                    newMergeGroups[i] = union(g1, g2)
+                    newMergeGroups[j] = []
+                    somethingChanged = true
+                }
+            }
+        }
+    }
```
jsoules, line 122: Ah, I see. You need to do union between any pair of merge groups which are not disjoint, then flatten the resulting groups, then remove any group that has fewer than 2 members (since it isn't a party if nobody comes).

src/extensions/extensionInterface.ts
jsoules: It's not immediately obvious what the intended behavior of the `UnmergeUnits` verb is.

Let's assume that we never get here without having a simplified (disjoint) merge group list as I think that'll be ensured by the way the `MergeUnit` verb is defined.

So, supposing I have merge groups like: `[[1, 2, 3, 5], [4, 6, 7], [8, 10]]`.

- If I call `UnmergeUnits` with `[1]`, I should get back `[[2, 3, 5], [4, 6, 7], [8, 10]]`, right? Or does it dissolve the group containing `1` entirely?
- If I call with `[1, 2]`, assuming the first merge group becomes `[3, 5]`--do `[1, 2]` remain grouped together?
- If I call with `[3, 4]`, which are not in the same group, do I wind up with `[[1, 2, 5], [6, 7]...]` having removed these items from their respective groups? Or does the unmerge fail, because units 3 and 4 weren't merged to each other?

I think it'd be helpful to document the intended behavior in comments here because we could get into some confusing situations.

src/extensions/mountainview/MVSortingView/CurationControl.tsx
jsoules: Nice to see custom labels back!

src/extensions/extensionInterface.ts
jsoules: (Even if the UI looks like it disallows some of these situations from occurring)

src/extensions/unitstable/Units/UnitsTable.tsx
jsoules: I think it might actually be easier to understand at a glance if you give the full set of merged units in the parens, rather than omitting that row's unit ID--that would make it more immediately obvious in this instance that all three of the selected items are merged into the same group.

src/extensions/unitstable/Units/UnitsTable.tsx
```diff
@@ -130,7 +130,11 @@ const UnitsTable: FunctionComponent<Props> = (props) => {
...
+        const mergeGroups = (sorting.curation || {}).mergeGroups || []
+        return mergeGroups.filter(g => (g.includes(unitId)))[0] || null
```
jsoules, line 136: Assumes that a unit will appear in exactly one or zero merge groups. (Probably we will have maintained the list of merge groups so that it has this property--but that's emergent behavior from the way we're maintaining the list, rather than an enforced guarantee, so I wanted to note it.)

jsoules (commented):
Had a few suggestions and a few questions in this one. There's a lot of subtleties in these operations so I think it's worth it to talk through the edge cases and provide a little bit more in the way of documentation/comments than we would otherwise need to.

src/extensions/extensionInterface.ts
jsoules: I wonder if the merge/unmerge operations would be simplified by representing the merge groups as an object whose keys are the unit IDs and whose values are Sets. Initially each object points to a Set that contains only itself. On merge, use the smallest unit ID as the key and add the full set of merged units into this Set, then update the object pointed to by all the other unit IDs to be that one.

So if we have `merge(group1 = [1, 3, 5], group2 = [2, 4, 7])` just do:
```
const finalSet = mergeGroups(group1[0]) // they all point to the same object so any one of them will do

for (const item of group2 {
  finalSet.add(item) // redundant adds are ignored
  mergeGroups[item] = finalSet // update group2's units' references to point to the final set, which will be mutated in place
}
```

And then to unmerge:
```
for (const x in items_to_unmerge) {
  mergeGroups[x].remove(x)
  mergeGroups[x] = new Set([x])
}
```

That said, what you've already written is already in place and working, and I don't know about JSON serializability (if we'd need to convert the Sets back to lists or something). Initial state is easy when nothing is merged, but loading a curation that has some merges in it would be non-obvious as you'd have to compare the lists and make sure they all pointed to the correct Set reference (rather than initializing with multiple Sets that happen to have the same values).

src/containers/SortingView.tsx
magland: i've been avoiding the `?.` because I got compiler errors when trying to compile the jupyter extension. However, I think I've resolved that issue now, and so my avoidance of `?.` is purely a habit right now. Will revisit this.

src/extensions/extensionInterface.ts
magland: The simplifyMergeGroups ensures that merge groups are disjoint. So when dispatch actions are applied, it should address this issue. Yes mergeGroupForUnit(unitId) should be a convenience function that we build into the interface. Making an issue for this.

src/extensions/extensionInterface.ts
magland: The behavior is: remove unit(s) from all merge groups and then simplify. Adding an issue to document more clearly.

src/extensions/unitstable/Units/UnitsTable.tsx
magland: agree. Making an issue for this.

src/extensions/extensionInterface.ts
magland: while there are advantages to what you propose in terms of ease of manipulation of the merge groups, and access, I think I prefer to have a simpler representation (sets of disjoint sets) and then create convenience functions for doing the manipulations. We're not worried about efficiency here.

magland:
TODO: Add docs to explain the choice for merge group representation as lists of disjoint lists of unit IDs

PR #217, opened 2020-12-18
jupyter: import external unit metrics

magland:
Allow importing of external unit metrics from jupyter notebook as follows:

image

### DIFF
--- a/jupyterlab/labbox_ephys_widgets_jp/labbox_ephys_widgets_jp/create_sorting_view.py
+++ b/jupyterlab/labbox_ephys_widgets_jp/labbox_ephys_widgets_jp/create_sorting_view.py
@@ -9,6 +9,7 @@ import numpy as np
 from hither.job import Job as HitherJob
 from ipywidgets import DOMWidget
 from traitlets import Dict as DictTrait
+from traitlets import List as ListTrait
 from traitlets import Unicode
 
 from ._frontend import module_name, module_version
@@ -39,6 +40,7 @@ def create_sorting_view(plugin_name: str, *, sorting: le.LabboxEphysSortingExtra
         sortingInfo = DictTrait(le.get_sorting_info(sorting_object=sorting.object(), recording_object=recording.object())).tag(sync=True)
         selection = DictTrait({}).tag(sync=True)
         curation = DictTrait({}).tag(sync=True)
+        externalUnitMetrics = ListTrait([]).tag(sync=True)
         def __init__(self) -> None:
             super().__init__()
             self._jobs_by_id: Dict[str, HitherJob] = {}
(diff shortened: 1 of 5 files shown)


jupyterlab/labbox_ephys_widgets_jp/labbox_ephys_widgets_jp/create_sorting_view.py
```diff
@@ -39,6 +40,7 @@ class SortingView(DOMWidget):
         sortingInfo = DictTrait(le.get_sorting_info(sorting_object=sorting.object(), recording_object=recording.object())).tag(sync=True)
         selection = DictTrait({}).tag(sync=True)
         curation = DictTrait({}).tag(sync=True)
+        externalUnitMetrics = ListTrait([]).tag(sync=True)
```
magland, line 43: externalUnitMetrics is now an attribute of the sorting view in jupyter. You should be able to set it either before or after the widget is created.

jupyterlab/labbox_ephys_widgets_jp/package.json
```diff
@@ -11,6 +11,8 @@
   "files": [
     "lib/**/*.js",
     "dist/*.js",
+    "lib/**/*.css",
```
magland, line 14: Had some trouble with the new css files during building of the jupyterlab extension

jupyterlab/labbox_ephys_widgets_jp/package.json
```diff
@@ -31,7 +33,7 @@
   "scripts": {
     "build": "npm run build:lib && npm run build:nbextension",
     "build:labextension": "npm run clean:labextension && mkdirp labbox_ephys_widgets_jp/labextension && cd labbox_ephys_widgets_jp/labextension && npm pack ../..",
-    "build:lib": "tsc",
+    "build:lib": "tsc && npm run copy-files",
```
magland, line 36: Turns out that tsc only transpiles ts -> js and does not copy other files (such as .css) In order to handle css properly with the jupyterlab extension, we need to have a step that copies those css files into the lib/ directory. See the copy-files command below.

jupyterlab/labbox_ephys_widgets_jp/src/pluginWidgets.tsx
```diff
@@ -249,8 +250,24 @@ const PluginComponentWrapper: FunctionComponent<PluginComponentWrapperProps> = (
       selection: model.get('selection')
     })
   }, null)
+
+  // externalUnitMetrics
```
magland, line 254: We handle the externalUnitMetrics attribute in a very similar way to curation and selection attributes.

src/extensions/extensionInterface.ts
```diff
@@ -172,6 +172,20 @@ export const sortingCurationReducer: Reducer<SortingCuration, SortingCurationAct
 }
 ////////////////////
 
+// This reducer is used only by the jupyter extension
```
magland, line 175: A new reducer for use in the jupyter extension

jsoules (commented):
All seems reasonable.

Issue #143, opened 2020-12-04
Test unit selection on mac


magland:
Apparently Ctrl+click may not work the same (according to Kyu)

jsoules:
Finally broke out the macbook and gave this a shot.

It looks like ctrl-click is intercepted by the browser to interpret as right-click/context menu.

I'm not sure there's anything we can do about this, but it is an annoyance for mac users. That said, it's also pretty much a UI standard in the Windows and Linux world, so I wouldn't do away with it entirely.

Two points:
1) This is less important with the ever-present checkboxes in the new layouts.
2) We could try mapping the same functionality to the *meta* key, which maps to the command key on most Macs. (Some Linux users may have a meta key mapped, but they probably won't notice getting an additional way to do things.)
3) We could also map the ctrl-click functionality for this selection effect so that it also fires if the alt key is pressed (I already have the `alt` modifier key being captured by the framework in [`formClickEventFromMouseEvent`](https://github.com/laboratorybox/labbox-ephys/blob/498795d486b69309b19aa93ca0296a340c58db36/src/extensions/CanvasWidget/CanvasWidgetLayer.ts#L86-L99).

I think the path of least resistance is to allow the meta key (if defined) to act as a synonym for the control key. Submitting a PR to that effect (it's a one-line change, but I haven't actually tested it because the macbook is still installing updates...)

jsoules:
Should be fixed by PR #221.

PR #221, opened 2020-12-18
(Fixes Issue #143) Add meta key modifier as an alias for control key modifier.

jsoules:
That's it, that's the whole PR.

I just wanted to go through the formal process because while I've tested that it doesn't break things on Ubuntu, my Mac is still installing several months of updates, so I won't be able to verify that this fixes Kyu's problem for another half hour or so.


magland:
nice one. feel free to merge when you are ready

PR #235, opened 2020-12-22
units table sorting and improved appearance

magland:


.vscode/labbox-snippets.code-snippets
```diff
@@ -15,11 +15,10 @@
 	// 	],
 	// 	"description": "Log output to console"
 	// }
-	"Print to console": {
-		"scope": "javascript,typescript",
-		"prefix": "debug",
+	"consolelog": {
```
magland, line 18: adjust console.log code snippet

devel/example_notebooks/example1.ipynb
```diff
@@ -73,6 +73,7 @@
    "source": [
     "# You can manipulate the curation via python and then set the mutated object back to the GUI\n",
     "# Note: this doesn't seem to be working at the moment\n",
+    "if 'labelsByUnit' not in curation: curation['labelsByUnit'] = {}\n",
```
magland, line 76: add this line to notebook example in case no curation was performed on widget

jupyterlab/labbox_ephys_widgets_jp/labbox_ephys_widgets_jp/create_recording_view.py
```diff
@@ -50,15 +49,15 @@ def _handle_message(self, widget, msg, buffers):
...
-                    self.send(dict(type='test 3'))
                     self.send({
                         'type': 'hitherJobCreationError',
                         'client_job_id': client_job_id,
                         'error': str(err) + ' (new method)'
                     })
                     return
-                self.send(dict(type='test 4'))
-                setattr(job, '_client_job_id', client_job_id)
+                client_job_ids = getattr(job, '_client_job_ids', [])
```
magland, line 58: attach more than one client job id to the same hither job

package.json
```diff
@@ -2,6 +2,10 @@
   "name": "labbox-ephys",
   "version": "0.4.5",
   "dependencies": {
+    "@fortawesome/fontawesome-svg-core": "^1.2.32",
```
magland, line 5: font awesome icons

src/containers/SortingView.tsx
```diff
@@ -170,8 +170,26 @@ const SortingView: React.FunctionComponent<Props> = (props) => {
   //   'paddingLeft': '20px'
   // }
 
-  const contentWrapperStyle = {
-    'marginLeft': 0
+  const W = props.width || 800
```
magland, line 173: improve layout of SortingView

src/extensions/electrodegeometry/electrodegeometry.tsx
```diff
@@ -1,6 +1,8 @@
 // LABBOX-EXTENSION: electrodegeometry
 // LABBOX-EXTENSION-TAGS: jupyter
 
+import { faThLarge } from '@fortawesome/free-solid-svg-icons';
```
magland, line 4: Put in some icons

src/extensions/extensionInterface.ts
```diff
@@ -48,6 +48,7 @@ export interface LabboxPlugin {
     priority?: number
     disabled?: boolean
     development?: boolean
+    icon?: JSX.Element
```
magland, line 51: Allow plugins to provide icons.

src/extensions/unitstable/Units/TableWidget.tsx
```diff
@@ -0,0 +1,195 @@
+import { faCaretDown, faCaretUp } from '@fortawesome/free-solid-svg-icons';
```
magland, line 1: Split UnitsTable into two components. One is TableWidget, a more generic widget that can sort columns.

jupyterlab/labbox_ephys_widgets_jp/labbox_ephys_widgets_jp/create_recording_view.py
jsoules: So this is basically going to mean that multiple computations can be pushed to the remote server with one hither call, is that right?

magland: Well. It means that we can handle the situation where multiple client-side jobs correspond to the same server-side job. Actually don't know if this happens in practice. If it did happen, then it would be because the identical job was requested multiple times prior to the job completing. But again I don't know if it actually happens. I was having some trouble and thought this might be the cause.

PR #239, opened 2020-12-23
simplify hither calls

magland:


jupyterlab/labbox_ephys_widgets_jp/labbox_ephys_widgets_jp/create_recording_view.py
```diff
@@ -13,16 +13,25 @@
 
 from ._frontend import module_name, module_version
```
magland, line 14: Avoid code duplication by using the WorkerSession from GUI for handling hither calls in jupyter widgets.

jupyterlab/labbox_ephys_widgets_jp/labbox_ephys_widgets_jp/create_recording_view.py
```diff
@@ -37,101 +46,15 @@ class RecordingView(DOMWidget):
         recordingInfo = DictTrait(le.get_recording_info(recording_object=recording.object())).tag(sync=True)
         def __init__(self) -> None:
             super().__init__()
-            self._jobs_by_id: Dict[str, HitherJob] = {}
             self.on_msg(self._handle_message)
+            self._worker_session = le.WorkerSession(labbox_config=labbox_config)
```
magland, line 50: Here is where we create a worker session (the one used for the web gui)

python/labbox_ephys/api/_workersession.py
```diff
@@ -99,18 +99,31 @@ def handle_message(self, msg):
...
-                    job = hi.run(functionName, **kwargs, labbox=self._labbox_context).wait()
-                except Exception as err:
-                    self._send_message({
-                        'type': 'hitherJobCreationError',
-                        'client_job_id': client_job_id,
-                        'error': str(err) + ' (new method)'
-                    })
-                    return
+            try:
+                outer_job = hi.run(functionName, **kwargs, labbox=self._labbox_context)
```
magland, line 104: This is the outer job... i.e. the hither job that creates and return a hither job to be queued.

python/labbox_ephys/api/_workersession.py
```diff
@@ -99,18 +99,31 @@ def handle_message(self, msg):
...
+            except Exception as err:
+                self._send_message({
```
magland, line 106: In this case there was an error running the outer job

python/labbox_ephys/api/_workersession.py
```diff
@@ -99,18 +99,31 @@ def handle_message(self, msg):
...
+                    'type': 'hitherJobError',
+                    'job_id': client_job_id,
+                    'client_job_id': client_job_id,
+                    'error_message': f'Error creating outer job: {str(err)}',
+                    'runtime_info': None
+                })
+                return
+            try:
+                job_or_result = outer_job.wait()
```
magland, line 115: We wait on the outer job to run. It will either return a hither job (to be queued) or an immediate result. If an immediate result, then we send back saying the job has finished. Otherwise, we will queue that job and send the finished message later.

src/actions/getRecordingInfo.ts
```diff
@@ -6,8 +6,7 @@ export const getRecordingInfo = async (a: {recordingObject: any, hither: HitherC
         'createjob_get_recording_info',
         { recording_object: a.recordingObject },
         {
-            useClientCache: true,
-            newHitherJobMethod: true
+            useClientCache: true
```
magland, line 9: Simplify hither calls throughout. The only option on the client (javascript) side is useClientCache which specifies whether we cache the result in the browser memory.

src/extensions/common/PlotGrid.tsx
```diff
@@ -2,6 +2,7 @@ import { Button, Grid } from '@material-ui/core';
 import React, { FunctionComponent, useState } from 'react';
 import { CalculationPool, HitherContext, Sorting } from '../extensionInterface';
 import ClientSidePlot from './ClientSidePlot';
+import './localStyles.css';
```
magland, line 5: I needed to move localStyles.css so that it would apply the jupyter widgets as well as the web gui.

src/containers/SortingView.tsx
```diff
@@ -66,7 +66,7 @@ const SortingView: React.FunctionComponent<Props> = (props) => {
   //     const sortingInfo = await createHitherJob(
   //       'createjob_get_sorting_info',
   //       { sorting_object: sorting.sortingObject, recording_object: recording.recordingObject },
-  //       { kachery_config: {}, useClientCache: true, wait: true, newHitherJobMethod: true}
+  //       { kachery_config: {}, useClientCache: true }
```
jsoules, line 69: Nice not to have to specify the `newHitherJobMethod` everywhere too.

src/extensions/common/PlotGrid.tsx
jsoules: Makes sense.

python/labbox_ephys/api/_workersession.py
magland: exactly.

src/containers/ImportSortings.tsx
```diff
@@ -154,17 +154,15 @@ const ImportSortingFromSpikeForest: FunctionComponent<{ onDone: () => void, onAd
                         recording_object: recordingObject
                     },
                     {
-                        useClientCache: false
+                        useClientCache: false // todo: maybe this should be true?
```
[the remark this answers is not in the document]
magland: we'll need to address garbage-collection at all levels, including job cache on the server side (disk) and on the client side (browser RAM). For now it's not causing problems for me at least, but that may change.

src/containers/SortingView.tsx
magland: yes... everything is now ported to new hither job method.

PR #242, opened 2020-12-23
fix cross correlograms view

magland:


src/actions/hitherJobs.ts
```diff
@@ -7,8 +7,8 @@ export const addHitherJob = (job: HitherJob): RootAction => ({
     job: job
 })
 
-export const updateHitherJob = (args: { jobId: string, update: HitherJobUpdate}): RootAction => ({
+export const updateHitherJob = (args: { clientJobId: string, update: HitherJobUpdate}): RootAction => ({
```
magland, line 10: Use clientJobId rather than the (python) jobId to monitor jobs on the client side. This is because if the job fails to create on the python side, it will not have a job ID, only a client job ID.

src/extensions/correlograms/CrossCorrelogramsView/CrossCorrelogramsView.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { FunctionComponent } from 'react'
```
magland, line 1: Here is the new cross correlograms view, modeled after the spike amplitudes view where the unit selection is in the left panel, with a splitter.

src/extensions/correlograms/CrossCorrelogramsView/CrossCorrelogramsWidget.tsx
```diff
@@ -1,45 +1,41 @@
-import { Button, Grid } from '@material-ui/core';
-import React, { useState } from 'react';
-import { SizeMeProps, withSize } from 'react-sizeme';
-import ClientSidePlot from '../common/ClientSidePlot';
-import createCalculationPool from '../common/createCalculationPool';
-import { SortingViewProps } from '../extensionInterface';
-import Correlogram_rv from './Correlogram_ReactVis';
+import { Grid } from '@material-ui/core';
```
magland, line 1: The new cross correlograms widget, modified from CrossCorrelograms.tsx

src/extensions/correlograms/CrossCorrelogramsView/CrossCorrelogramsWidget.tsx
```diff
@@ -1,45 +1,41 @@
...
         // (the user can scroll). So: horizontal space taken is:
         // width = n*plotWidth + 2*margin (2 outer margins) + (n-1)*margin (gutters between plots)
         // width = margin*(n+1) + plotWidth * n
         // Solve for plotWidth = (width - margin*(n+1))/n.
         // And we can't have fractional pixels, so round down.
-        const plotWidth = Math.min(maxSize, Math.floor((width - marginInPx*(N + 1))/N));
-        return plotWidth;
+        const xPlotWidth = Math.floor((width - marginInPx*(N + 1))/N)
+        const yPlotWidth = Math.floor((height - marginInPx*(N + 1))/N)
+        return Math.max(minSize, Math.min(maxSize, Math.min(xPlotWidth, yPlotWidth)))
```
magland, line 32: Take into consideration the height when computing the plot size

src/extensions/correlograms/CrossCorrelogramsView/CrossCorrelogramsWidget.tsx
```diff
@@ -54,24 +50,22 @@ const CrossCorrelograms: React.FunctionComponent<SortingViewProps & SizeMeProps>
...
+                                    dataFunctionName='createjob_fetch_correlogram_plot_data'
                                     dataFunctionArgs={{
                                         sorting_object: sorting.sortingObject,
                                         unit_x: pair.xkey,
                                         unit_y: pair.ykey
                                     }}
                                     boxSize={{
                                         width: plotWidth,
-                                        height: plotWidth
+                                        height: plotWidth - 35
```
magland, line 68: We'll need to figure out how to do the plot box sizing in a better way. Here the "-35" is referring to vertical space taken up by labels. Will need to improve.

PR #252, opened 2020-12-26
jupyter: auto width/height for widgets

magland:


jupyterlab/labbox_ephys_widgets_jp/src/pluginWidgets.tsx
```diff
@@ -272,6 +272,33 @@ const PluginComponentWrapper: FunctionComponent<PluginComponentWrapperProps> = (
     }, null)
   }, [model])
 
+  const [divElement, setDivElement] = useState<HTMLDivElement | null>(null)
```
magland, line 275: Not ideal, but we do polling to determine what the size of the widget is.

jupyterlab/labbox_ephys_widgets_jp/src/pluginWidgets.tsx
```diff
@@ -340,9 +392,14 @@ export class SortingView extends DOMWidgetView {
...
   }
   render() {
-    // this.el.classList.add('custom-widget');
+    const pluginName = this.model.get('pluginName')
+    const plugin = extensionContext._sortingViewPlugins[pluginName]
+    if (!plugin) throw Error(`Plugin not found: ${pluginName}`)
 
+    this.el.classList.add('plugin-' + pluginName)
     const x = this.element()
+    this.el.style.height = '100%'
```
magland, line 401: height=100% is for when the widget is in it's own window (separate output tab in jupyterlab)
We use min-height for regular output cells. notebookCellHeight is now an optional parameter of the plugin.

PR #258, opened 2020-12-26
problem: fix unit click select issue

magland:


src/containers/SortingView.tsx
```diff
@@ -220,7 +192,6 @@ const SortingView: React.FunctionComponent<Props> = (props) => {
             recording={recording}
             selection={selection}
             selectionDispatch={selectionDispatch}
```
magland, line 194: Use selectionDispatch rather than onUnitClicked here.

src/extensions/averagewaveforms/AverageWaveformsView/AverageWaveformsView.tsx
```diff
@@ -1,18 +1,26 @@
...
 import { SortingViewProps } from "../../extensionInterface";
 import AverageWaveformView from './AverageWaveformView';
 
 const calculationPool = createCalculationPool({maxSimultaneous: 6})
 
-const AverageWaveformsView: FunctionComponent<SortingViewProps> = ({recording, sorting, selection, onUnitClicked, hither}) => {
+const AverageWaveformsView: FunctionComponent<SortingViewProps> = ({recording, sorting, selection, selectionDispatch, hither}) => {
     const selectedUnitIdsLookup: {[key: string]: boolean} = (selection.selectedUnitIds || []).reduce((m, uid) => {m[uid + ''] = true; return m}, {} as {[key: string]: boolean})
+    const handleUnitClicked = useCallback((unitId: number, event: {ctrlKey?: boolean, shiftKey?: boolean}) => {
+        selectionDispatch({
```
magland, line 12: Dispatch a unit clicked event rather than handling the unit click it here

src/extensions/extensionInterface.ts
```diff
@@ -211,7 +211,39 @@ type SetVisibleUnitIdsSortingSelectionAction = {
...
+type UnitClickedSortingSelectionAction = {
+    type: 'UnitClicked'
+    unitId: number
+    ctrlKey?: boolean
+    shiftKey?: boolean
+}
+
+export type SortingSelectionAction = SetSelectionSortingSelectionAction | SetSelectedUnitIdsSortingSelectionAction | SetVisibleUnitIdsSortingSelectionAction | UnitClickedSortingSelectionAction
+
+const unitClickedReducer = (state: SortingSelection, action: UnitClickedSortingSelectionAction): SortingSelection => {
```
magland, line 223: A reducer for handling unit clicks, with optional ctrl or shift keys

src/extensions/extensionInterface.ts
```diff
@@ -211,7 +211,39 @@ type SetVisibleUnitIdsSortingSelectionAction = {
...
+            }
+        }
+        else {
+            return {
+                ...state,
+                selectedUnitIds: [...(state.selectedUnitIds || []), unitId]
+            }
+        }
+    }
+    // todo: restore anchor/shift-select behavior somewhere
```
magland, line 239: Will need to restore shift-click range select functionality somewhere. (maybe not in the reducer because we need to know the list of unit ids)

src/extensions/old/averagewaveformsold/AverageWaveformsNew.tsx
```diff
@@ -22,11 +22,19 @@ const AverageWaveformsNew: React.FunctionComponent<SortingViewProps> = (props) =
...
     // return <div>AAAA: {calcMode} {result + ''}</div>
     const selectedUnitIdsLookup: {[key: string]: boolean} = (selection.selectedUnitIds || []).reduce((m, uid) => {m[uid + ''] = true; return m}, {} as {[key: string]: boolean})
+    const handleUnitClicked = useCallback((unitId: number, event: {ctrlKey?: boolean, shiftKey?: boolean}) => {
+        selectionDispatch({
+            type: 'UnitClicked',
+            unitId,
+            ctrlKey: event.ctrlKey,
+            shiftKey: event.shiftKey
+        })
+    }, [selectionDispatch])
```
jsoules, line 32: I'd be nice to centralize this a bit more if it's possible to do so, I see it's repeated over a couple different files.

Actually I'm not too clear on the distinction between `src/extensions/old` and `src/extensions/` and what is & isn't current. Some of this may have been superseded by later PRs. Are we due for a little New Year cleaning in terms of unused files & commented code?

src/extensions/extensionInterface.ts
jsoules: Hmmm... yeah, since they can be discontinuous, without knowing the exact set of unit IDs it can be annoying to try to handle range select. I do recall at some point or another I handled this by just throwing the entire newly selected range into the set of selected Unit IDs and letting the front-end figure out if the corresponding unit IDs even existed (...and of course I spent several hours tracking down bugs that resulted from referencing selected unit IDs that didn't exist when I forgot to handle it properly, so maybe that isn't a productive reminiscence).

Would it be at all practical to design the `unitClickedReducer` to store the state on the set of unit IDs (implying, for it to actually be initialized at some point with the full list of IDs that exist)?

As a separate thought on the topic of range selection, we had been handling discontinuous ranges of IDs but we were always assuming that the unit IDs would be displayed on screen in ascending order. If, for some reason, that was not the case, then the old range select would break anyway. (I don't know if this is at all likely or possible, but I consider thinking of implausible scenarios part of my job.) So if & when we reimplement range select, it might be beneficial to have a list that actually preserves the ordering of the unit IDs.

Actually I suppose unit IDs could very easily be reordered in the units table since we have sorting functionality; while selection in that context is handled by check boxes, I wonder if we shouldn't coordinate display in other views with whatever ordering is currently active in the units table... so that e.g. an autocorrelograms view would resort the charts if the user re-sorts the table?

src/extensions/old/averagewaveformsold/AverageWaveformsNew.tsx
magland: Yeah, maybe a `useHandleUnitClicked(selectionDispatch)` custom hook

The plan is for `old/` stuff to go away at some point, but keeping it in for now to avoid breaking and for reference (e.g., bokeh stuff).

magland:
@jsoules, I don't know why github is not allowing me to reply to your comment on extensionInterface.ts. I'll just reply here.

There are various possibilities, my current preference (though not committed) would be to pass an ordered list of the unit IDs to the dispatch event that does the selection. In other words, have this info as part of the action rather than part of the persisted state.

jsoules:
> There are various possibilities, my current preference (though not committed) would be to pass an ordered list of the unit IDs to the dispatch event that does the selection. In other words, have this info as part of the action rather than part of the persisted state.

That sounds like a smart approach.

PR #262, opened 2020-12-28
MVSortingView: add vertical splitter betw north/south

magland:


src/extensions/common/VSplitter.tsx
```diff
@@ -0,0 +1,125 @@
+import React, { FunctionComponent, ReactElement, useState } from 'react';
```
magland, line 1: I added this VSplitter class but didn't actually use it (instead made a custom solution). But I preserve this code for future use. The VSplitter.tsx is just a copy of Splitter.tsx with a few minor adjustments to make a vertical splitter rather than horizontal.

src/extensions/mountainview/MVSortingView/ViewContainer.tsx
```diff
@@ -18,6 +19,7 @@ const ViewContainer: FunctionComponent<Props> = ({ children, views, onViewClosed
     const [currentNorthView, setCurrentNorthView] = useState<View | null>(null)
     const [currentSouthView, setCurrentSouthView] = useState<View | null>(null)
     const [activeArea, setActiveArea] = useState<'north' | 'south'>('north')
+    const [splitterFrac, setSplitterFrac] = useState(0.5)
```
magland, line 22: Keep track of the splitter frac (Rather than the absolute possition) in order to work well with window resizes.

src/extensions/mountainview/MVSortingView/ViewContainer.tsx
```diff
@@ -40,29 +42,40 @@ const ViewContainer: FunctionComponent<Props> = ({ children, views, onViewClosed
...
     const W = (width || 300) - hMargin * 2
     const H = height - vMargin * 2
+    const splitterHeight = 16
+
+    const a = (H - tabBarHeight * 2 - splitterHeight)
+    const H1 = a * splitterFrac
+    const H2 = a * (1 - splitterFrac)
+
+    const handleSplitterDelta = useCallback((delta: number) => {
+        setSplitterFrac((H1 + delta) / (H1 + H2))
```
magland, line 52: The splitter grip has been moved, so we calculate the new splitter fraction

src/extensions/common/VSplitter.tsx
```diff
@@ -0,0 +1,125 @@
...
+    let child2 = children[1] as any as ReactElement
+
+    if (child2 === undefined) {
+        return <child1.type {...child1.props} width={width} height={height} />
+    }
+
+    const gripWidth = props.gripWidth ?? 12
+    const gripInnerWidth = props.gripInnerWidth ?? 4
+    const gripMargin = props.gripMargin ?? 4
+    const height1 = gripPosition - gripWidth / 2 - gripMargin
```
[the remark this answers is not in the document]
magland: I thought about calling in height, but it's width of the grip line... so it's not width in the meaning of horizontal width... it's like pen width. Really thickness is the word. It should be called thickness.

PR #266, opened 2020-12-29
implement recording selection and sync selected electrode ids across views

magland:


magland:
Also syncing amp scale factor across views

python/labbox_ephys/misc/get_recording_info.py
```diff
@@ -6,18 +6,28 @@
...
-@hi.function('createjob_get_recording_info', '0.1.1')
+@hi.function('createjob_get_recording_info', '0.1.2')
 def createjob_get_recording_info(labbox, recording_object):
     jc = labbox.get_default_job_cache()
     with hi.Config(
         job_cache=jc
     ):
         return get_recording_info.run(recording_object=recording_object)
 
+def estimate_noise_level(recording: se.RecordingExtractor):
```
magland, line 20: We now estimate the noise level as part getting the recording info

python/labbox_ephys/misc/get_recording_info.py
```diff
@@ -6,18 +6,28 @@
...
+    N = recording.get_num_frames()
+    samplerate = recording.get_sampling_frequency()
+    start_frame = 0
+    end_frame = int(np.minimum(samplerate * 1, N))
+    X = recording.get_traces(channel_ids=[int(id) for id in recording.get_channel_ids()], start_frame=start_frame, end_frame=end_frame)
```
magland, line 25: Estimate the noise level based on the first 1 sec of data

src/extensions/common/CanvasWidget/CanvasPainter.ts
```diff
@@ -129,11 +129,16 @@ export class CanvasPainter {
         }
     }
     useOffscreenCanvas(W: number, H: number) {
-        const c = new OffscreenCanvas(W, H)
-        this._offscreenCanvas = c
-        const cc = c.getContext('2d')
-        if (!cc) throw Error('Unexpected')
-        this._context2D = cc
+        try {
```
magland, line 132: Be more careful about the OffscreenCanvas creation so that we don't crash the app

src/extensions/extensionInterface.ts
```diff
@@ -193,8 +193,88 @@ export const externalUnitMetricsReducer: Reducer<ExternalSortingUnitMetric[], Ex
 }
 ////////////////////////////////
 
+// Recording selection
```
magland, line 196: The recording selection includes selected electrode IDs, current timepoint, time range, and amplitude scale factor.

src/extensions/timeseries/TimeWidgetNew/TimeWidgetNew.tsx
```diff
@@ -197,13 +201,47 @@ const plotMargins = {
     bottom: 100
 }
 
+const useTimeState = (args: {
```
magland, line 204: This custom hook needs to be simplified. I was trying to achieve both controlled and uncontrolled state for the recording selection. Not very elegant.

jsoules: Sorry, I'm not sure I follow what you mean by 'controlled and uncontrolled state' in this context.

src/extensions/timeseries/TimeWidgetNew/TimeWidgetNew.tsx
```diff
@@ -197,13 +201,47 @@ const plotMargins = {
...
+}): {timeState: TimeState, timeDispatch: React.Dispatch<TimeAction>} => {
+    const { numTimepoints, maxTimeSpan, currentTimepoint, onCurrentTimepointChanged, timeRange, onTimeRangeChanged } = args
+    const [timeState, setTimeState] = useState<TimeState>({timeRange: null, currentTime: null, numTimepoints, maxTimeSpan})
+    let newTimeState = {...timeState}
+    if (currentTimepoint !== undefined) {
+        newTimeState = {...newTimeState, currentTime: currentTimepoint}
+    }
+    if (timeRange !== undefined) {
+        newTimeState = {...newTimeState, timeRange}
+    }
```
jsoules, line 220: Is there a circumstance under which `currentTimepoint` or `timeRange` could be reset to `undefined` when they previously had a value that could be preserved? I guess a particular call to `useTimeState` might omit those optional parameters if they aren't meant to be changed?

python/labbox_ephys/misc/get_recording_info.py
```diff
@@ -6,18 +6,28 @@
...
+    est_noise_level = np.median(np.abs(X.squeeze())) / 0.6745  # median absolute deviation (MAD) estimate of stdev
```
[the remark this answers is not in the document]
magland: I think you are right. I created an issue from it.

src/components/RecordingInfoView.js
```diff
@@ -29,7 +29,7 @@ const RecordingInfoView = ({ recordingInfo, hideElectrodeGeometry }) => {
                     num_frames={ri.num_frames}
                     channel_ids={ri.channel_ids}
                     channel_groups={ri.channel_groups}
-                    is_local={ri.is_local}
+                    noise_level={ri.noise_level}
```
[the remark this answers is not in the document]
magland: Yeah, it's not really a property of a recording. We want the recording info computation to be cached, but not the is_local. So I put this in a separate function that doesn't get cached (for purpose of checking whether we need to download during import)

src/extensions/common/PlotGrid.tsx
```diff
@@ -15,14 +15,15 @@ interface Props {
...
+    plotComponentPropsCallback?: any
     calculationPool: CalculationPool | undefined
     hither: HitherContext
 }
 
 const PlotGrid: FunctionComponent<Props> = ({ sorting, onUnitClicked, selections,
     dataFunctionName, dataFunctionArgsCallback, // fix this
     boxSize = { width: 200, height: 200},
-    plotComponent, plotComponentArgsCallback, // fix this
+    plotComponent, plotComponentArgsCallback, plotComponentPropsCallback, // fix this
```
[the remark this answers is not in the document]
magland: PlotGrid is being replaced... see the newest Autocor and Cross-cor views

src/extensions/timeseries/TimeseriesViewNew/TimeseriesViewNew.tsx
```diff
@@ -113,6 +116,12 @@ const TimeseriesViewNew = (props: Props) => {
     }
     useEffect(() => { effect(); });
 
+    const recordingSelection = props.recordingSelection || recordingSelectionInternal
+    const recordingSelectionDispatch = props.recordingSelectionDispatch || recordingSelectionInternalDispatch
```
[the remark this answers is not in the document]
magland: basically yes. We want some props to optionally be controlled or not controlled by the parent component. I couldn't find info on the preferred way to do that. Still planning to search for a clearer solution.

src/extensions/timeseries/TimeseriesViewNew/TimeseriesWidgetNew.tsx
```diff
@@ -229,15 +217,12 @@ const TimeseriesWidgetNew = (props: Props) => {
...
-        if (yScaleState.yScale !== prevYScale) {
-            if (panels) {
-                panels.forEach(p => {
-                    p.setYScale(yScaleState.yScale)
-                })
-            }
-            setPrevYScale(yScaleState.yScale)
+        if (panels) {
+            panels.forEach(p => {
+                p.setYScale(recordingSelection.ampScaleFactor || 1)
```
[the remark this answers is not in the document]
magland: ampScaleFactor is now synchronized across all the views. It is in the recording selection state (just like current timepoint). Not sure that this is always desirable, but so far it has proven to be very convenient.

src/extensions/timeseries/TimeWidgetNew/TimeWidgetNew.tsx
magland: I think we can assume this woud not change because the parent child component relationship should be consistent for this. But I think handling this could certainly be improved.

PR #270, opened 2020-12-29
Implement time animation and begin firetrack view

magland:


jupyterlab/labbox_ephys_widgets_jp/src/pluginWidgets.tsx
```diff
@@ -257,6 +257,7 @@ const PluginComponentWrapper: FunctionComponent<PluginComponentWrapperProps> = (
 
   // selection
   const [selection, selectionDispatch] = useReducer(sortingSelectionReducer, model.get('selection').selectedUnitIds ? model.get('selection') : {})
+  useRecordingAnimation(selectionDispatch)
```
magland, line 260: This special hook starts the animation.

src/extensions/extensionInterface.ts
```diff
@@ -199,6 +199,23 @@ export interface RecordingSelection {
...
     timeRange?: {min: number, max: number} | null
     ampScaleFactor?: number
+    animation?: {
+        lastUpdateTimestamp: number
+        currentTimepointVelocity: number // timepoints per second
+    }
+}
+
+export const useRecordingAnimation = (selectionDispatch: RecordingSelectionDispatch) => {
+    const [animationFrame, setAnimationFrame] = useState(0)
```
magland, line 209: Dispatch the AnimateRecording action every 100 ms

src/extensions/extensionInterface.ts
```diff
@@ -233,7 +250,16 @@ type ScaleAmpScaleFactorRecordingSelectionAction = {
     multiplier: number
 }
 
-export type RecordingSelectionAction = SetRecordingSelectionRecordingSelectionAction | SetSelectedElectrodeIdsRecordingSelectionAction | SetCurrentTimepointRecordingSelectionAction | SetTimeRangeRecordingSelectionAction | SetAmpScaleFactorRecordingSelectionAction | ScaleAmpScaleFactorRecordingSelectionAction
+type SetCurrentTimepointVelocityRecordingSelectionAction = {
+    type: 'SetCurrentTimepointVelocity',
+    velocity: number // timepoints per second
```
magland, line 255: This is how many timepoints we advance each second

src/extensions/extensionInterface.ts
```diff
@@ -269,6 +295,45 @@ export const recordingSelectionReducer: Reducer<RecordingSelection, RecordingSel
...
+            ...state,
+            animation: {
+                ...(state.animation || {}),
+                lastUpdateTimestamp: Number(new Date()),
+                currentTimepointVelocity: action.velocity
+            }
+        }
+    }
+    else if (action.type === 'AnimateRecording') {
+        const lastUpdate = state?.animation?.lastUpdateTimestamp || Number(new Date())
```
magland, line 309: Here is where the animation happens. Basically, we advance the current timepoint by an amount dependent on the velocity and how much time has elapsed since the last update.

src/extensions/firetrack/FireTrackView/FireTrackView.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { FunctionComponent, useCallback } from 'react'
```
magland, line 1: Here we make a plugin view that allows us to start/stop the animation.

src/extensions/extensionInterface.ts
```diff
@@ -233,7 +250,16 @@ type ScaleAmpScaleFactorRecordingSelectionAction = {
...
+type SetCurrentTimepointVelocityRecordingSelectionAction = {
+    type: 'SetCurrentTimepointVelocity',
+    velocity: number // timepoints per second
+}
+
+type AnimateRecordingSelectionAction = {
+    type: 'AnimateRecording'
+}
+
+export type RecordingSelectionAction = SetRecordingSelectionRecordingSelectionAction | SetSelectedElectrodeIdsRecordingSelectionAction | SetCurrentTimepointRecordingSelectionAction | SetTimeRangeRecordingSelectionAction | SetAmpScaleFactorRecordingSelectionAction | ScaleAmpScaleFactorRecordingSelectionAction | SetCurrentTimepointVelocityRecordingSelectionAction | AnimateRecordingSelectionAction
```
jsoules, line 262: Clearly a lot of the recent changes over the last couple weeks have had to do with putting together the `RecordingSelection` object and associated Reducer; this seems like a good potential jumping-off point, or at least something worth featuring prominently in new documentation.

src/extensions/extensionInterface.ts
```diff
@@ -269,6 +295,45 @@ export const recordingSelectionReducer: Reducer<RecordingSelection, RecordingSel
...
+    else if (action.type === 'AnimateRecording') {
+        const lastUpdate = state?.animation?.lastUpdateTimestamp || Number(new Date())
+        const current = Number(new Date())
+        const elapsed = current - lastUpdate
+        if (elapsed !== 0) {
+            let state2 = {...state}
+            const currentTimepointVelocity = state?.animation?.currentTimepointVelocity || 0
+            const currentTimepoint = state.currentTimepoint
+            let somethingChanged = false
+            if ((currentTimepointVelocity) && (currentTimepoint !== undefined)) {
```
jsoules, line 317: This is a little confusing--`currentTimepointVelocity` has a default zero value. So is this part of the check making sure that the velocity is nonzero, i.e. we're going to advance the animation? If that's the intent, it might be clearer to do the zero comparison explicitly.

PR #272, opened 2020-12-30
simplify timeseries data model

magland:


src/extensions/firetrack/FireTrackView/FireTrackView.tsx
```diff
@@ -1,13 +1,17 @@
...
-const FireTrackView: FunctionComponent<SortingViewProps> = ({recording, sorting, selection, selectionDispatch}) => {
+const FireTrackView: FunctionComponent<SortingViewProps> = ({recording, sorting, selection, selectionDispatch, hither}) => {
     const handleStopAnimation = useCallback(( args) => {
         selectionDispatch({type: 'SetCurrentTimepointVelocity', velocity: 0})
     }, [ selectionDispatch ])
     const handleStartAnimation = useCallback(( args) => {
         selectionDispatch({type: 'SetCurrentTimepointVelocity', velocity: 200})
     }, [ selectionDispatch ])
+
+    const timeseriesData = useTimeseriesData(recording.recordingObject, recording.recordingInfo, hither)
```
magland, line 14: Simplify the timeseries model system with this custom hook: useTimeseriesData(...)

src/extensions/timeseries/TimeseriesViewNew/useTimeseriesModel.ts
```diff
@@ -1,3 +1,8 @@
+import { useRef } from "react"
```
magland, line 1: Here's the new file where we define all the timeseries model functionality and create the custom hook: useTimeseriesModel

src/extensions/timeseries/TimeseriesViewNew/useTimeseriesModel.ts
```diff
@@ -150,4 +155,73 @@ class TimeseriesModelNew {
...
+      const num_channels = recordingInfo.channel_ids.length
+      const segment_size = Math.ceil(segment_size_times_num_channels / num_channels)
+      const model = new TimeseriesModelNew({
+          samplerate: recordingInfo.sampling_frequency,
+          num_channels,
+          num_timepoints: recordingInfo.num_frames,
+          segment_size
+      })
+      model.onRequestDataSegment((ds_factor: number, segment_num: number) => {
+          ;(async () => {
```
jsoules, line 187: What's the semicolon do?

src/extensions/timeseries/TimeseriesViewNew/useTimeseriesModel.ts
```diff
@@ -150,4 +155,73 @@ class TimeseriesModelNew {
...
+export interface TimeseriesData {
+  getChannelData: (ch: number, t1: number, t2: number, ds_factor: number) => number[]
+  requestChannelData: (ch: number, t1: number, t2: number, ds_factor: number) => void
+  numChannels: () => number
+  numTimepoints: () => number
+  getSampleRate: () => number
+  onDataSegmentSet: (handler: (ds_factor: number, t1: number, t2: number) => void) => void
+}
+
+const useTimeseriesData = (recordingObject: any, recordingInfo: RecordingInfo, hither: HitherContext): TimeseriesData | null => {
```
jsoules, line 169: So as I understand it, this is a custom hook that provides an interface to the time series data in a recording object. When called, the hook checks if a model exists and if the recording object & metadata (`recordingInfo`) are the same as the last call. If the model is defined and the ultimate data source is unchanged, just return the model persisted by the framework.

If the model does not exist, or if there's a new data source, then we define a new model of type `TimeseriesModelNew` which provides some data and some functionality. The model contains the sample rate, the channel count, a total frame count, and a segment size (presumably this is some sort of page size related to how much data can be downloaded from source at once, since it scales based on number of channels). The model also provides a function `onRequestDataSegment` which takes a downscaling factor and a segment number; when called, this function uses hither to asynchronously fetch the corresponding segment of the time series.

Did I miss anything?

src/extensions/timeseries/TimeseriesViewNew/useTimeseriesModel.ts
magland: It's just a good pattern to add a ";" to lines that start with "(". That's because if the previous line ends with an identifier (and not a semicolon), you will get a hard-to-decipher syntax error.

src/extensions/timeseries/TimeseriesViewNew/useTimeseriesModel.ts
magland: Sounds about right.

PR #273, opened 2020-12-30
firetrack: prelim implementation

magland:


src/extensions/electrodegeometry/ElectrodeGeometryWidget/ElectrodeGeometryWidget.tsx
```diff
@@ -1,34 +1,8 @@
...
-    LIGHT = 'rgb(228, 228, 228)',
-    DARK = 'rgb(32, 32, 32)'
-}
-
-export type Electrode = {
-    label: string,
-    x: number,
-    y: number
-}
+import { useLayer, useLayers } from "../../common/CanvasWidget/CanvasWidgetLayer"
```
magland, line 3: Separate out the electrodeGeometryLayer into its own .ts file so it can be reused.

src/extensions/electrodegeometry/ElectrodeGeometryWidget/electrodeGeometryLayer.ts
```diff
@@ -0,0 +1,266 @@
+import { norm } from "mathjs"
```
magland, line 1: Separate out the electrodeGeometryLayer into its own .ts file so it can be reused.

src/extensions/firetrack/FireTrackView/FireTrackView.tsx
```diff
@@ -1,76 +1,41 @@
-import { Table, TableBody, TableCell, TableHead, TableRow } from '@material-ui/core'
-import React, { FunctionComponent, useCallback, useEffect, useReducer, useState } from 'react'
-import { SortingSelection, SortingViewProps } from "../../extensionInterface"
-import useTimeseriesData, { TimeseriesData } from '../../timeseries/TimeseriesViewNew/useTimeseriesModel'
+import React, { FunctionComponent, useCallback } from 'react'
```
magland, line 1: Develop the firetrack view

src/extensions/firetrack/FireTrackView/FireTrackWidget.tsx
```diff
@@ -0,0 +1,83 @@
...
+        return { label: x + '', x: loc[0], y: loc[1], electrodeId: x }
+    })
+}
+
+const valToColor = (v: number) => {
+    if (v <= 0) return 'black'
+    if (v >= 1) return 'white'
+    const x = Math.floor(v * 255)
+    return `rgb(${x}, ${x}, ${x})`
+}
```
jsoules, line 27: Should this live with the `CanvasPainter` code that handles color strings?

magland: We need to create colormaps. A color map takes a value between 0 and 1 and returns a color. I wonder if we can grab the colormaps from matplotlib somehow: https://matplotlib.org/3.3.3/tutorials/colors/colormaps.html

PR #278, opened 2021-01-02
simplify hither calls

magland:


jupyterlab/labbox_ephys_widgets_jp/src/pluginWidgets.tsx
```diff
@@ -377,17 +378,18 @@ export class SortingView extends DOMWidgetView {
...
-        hither={hither}
-        sortingObject={sortingObject}
-        recordingObject={recordingObject}
-        sortingInfo={sortingInfo}
-        recordingInfo={recordingInfo}
-        plugins={filterPlugins(plugins)}
-        calculationPool={calculationPool}
-        model={this.model}
-      />
+      <HitherContext.Provider value={hither}>
```
magland, line 381: The hither call mechanism is simplified by using a context provider and a useContext() hook. With this new system, we don't need to pass the hither context in the props from parent->child components.

jupyterlab/labbox_ephys_widgets_jp/src/pluginWidgets.tsx
```diff
@@ -403,93 +405,92 @@ export class SortingView extends DOMWidgetView {
...
-        calculationPool={calculationPool}
-        recordingSelection={{}}
-        recordingSelectionDispatch={(a: RecordingSelectionAction) => {}}
-      />
-    )
-    return x
-  }
-  render() {
-    // this.el.classList.add('custom-widget');
+// export class RecordingViewModel extends DOMWidgetModel {
```
magland, line 408: Commenting out the recording views for now. Will focus on sorting view sfor now, and will restore the recording views later.

src/reducers/index.ts
```diff
@@ -61,6 +58,6 @@ const rootReducer = combineReducers({
     plugins
 })
 
-export type RootAction =  DatabaseConfigAction | DocumentInfoAction | HitherContextAction | HitherJobsAction | PersistingAction | RecordingsAction | ServerConnectionAction | ServerInfoAction | SortingsAction | ExtensionsConfigAction | PluginsAction
+export type RootAction =  DatabaseConfigAction | DocumentInfoAction | HitherJobsAction | PersistingAction | RecordingsAction | ServerConnectionAction | ServerInfoAction | SortingsAction | ExtensionsConfigAction | PluginsAction
```
magland, line 61: The hither context is no longer part of the redux store.

jupyterlab/labbox_ephys_widgets_jp/src/pluginWidgets.tsx
```diff
@@ -93,7 +93,9 @@ class HitherJobManager {
       status: '',
       timestampStarted: 0,
       timestampFinished: null,
-      wait: async () => {} // this will be replaced below
+      clientCancelled: false,
```
magland, line 96: clientCancelled field is needed for the useHitherJob hook

src/extensions/common/useHitherJob.ts
```diff
@@ -0,0 +1,107 @@
...
+                status: 'error',
+                functionName,
+                functionArgs,
+                hitherJobOpts,
+                job
+            })
+        })
+        return {result: undefined, job}
+    }
+    else if (state.status === 'running') {
```
jsoules, line 53: There's a return at the end of the previous conditional (l. 51), so it might be clearer if this starts a new conditional statement entirely rather than immediately jumping to else-if. This would also let you use a single line to throw the error when `state.job` is undefined, rather than having to repeat it in each branch.

I'd probably also lean toward a switch statement on `state.status`.

src/containers/ImportRecordings.tsx
```diff
@@ -1,10 +1,10 @@
-import React, { Dispatch, FunctionComponent, useState } from 'react';
+import React, { Dispatch, FunctionComponent, useContext, useState } from 'react';
```
jsoules, line 1: This `useContext` business seems really nice!

src/index.tsx
```diff
@@ -193,15 +195,21 @@ const waitForDocumentInfo = async () => {
...
 const content = (
   // <React.StrictMode> // there's an annoying error when strict mode is enabled. See for example: https://github.com/styled-components/styled-components/issues/2154 
-  <MuiThemeProvider theme={theme}>
-    <Provider store={theStore}>
-      <Router>
-        <AppContainer />
-      </Router>
-    </Provider>
-  </MuiThemeProvider>
+  <HitherContext.Provider value={hither}>
```
jsoules, line 204: And so here's where the actual hither context being used gets inserted, right?

src/extensions/common/useHitherJob.ts
magland: I agree, submitting an issue for this.

src/index.tsx
magland: That's right.

PR #282, opened 2021-01-04
fix sortings table so doesn't hang

magland:


src/containers/RecordingView.tsx
```diff
@@ -40,22 +40,23 @@ type Props = StateProps & DispatchProps & OwnProps & RouteComponentProps
...
-    const rec = recording;
-    if (!rec.recordingInfo) {
-      try {
-        const info = await getRecordingInfo({ recordingObject: rec.recordingObject, hither });
-        onSetRecordingInfo({ recordingId: rec.recordingId, recordingInfo: info });
-      }
-      catch (err) {
-        console.error(err);
-        return;
+  useEffect(() => {
```
magland, line 46: Cleaned up this effect. It should be improved more in the future by centralizing the getRecordingInfo via useRecordingInfo custom hook. Making an issue for this.

src/containers/SortingsTable.tsx
```diff
@@ -27,48 +28,36 @@ interface OwnProps {
...
-                                useClientCache: true
-                            }
-                        )
-                        info = await sortingInfoJob.wait() as SortingInfo;
-                        onSetSortingInfo({ sortingId: sor.sortingId, sortingInfo: info });
-                    }
-                    catch (err) {
-                        console.error(err);
-                        return;
+        for (const sor of sortings) {
```
magland, line 39: Cleaned up this useEffect to avoid excessive redundant hither job creation (was causing page hanging). The infosInProgress ref is necessary and seems to work, but there is probably a more elegant way of doing this.

PR #288, opened 2021-01-04
consolidate hither job code into extensions/common/hither/

magland:


magland:
This is just moving code around to put all the hither stuff in one place: src/extensions/common/hither/

@jsoules

PR #299, opened 2021-01-07
Fix issue 296 & add explicit Cython installation for dev container.

jsoules:


.devcontainer/Dockerfile
```diff
@@ -89,6 +89,7 @@ RUN apt-get update && \
 RUN pip install ipython jupyter jupyterlab
 RUN pip install python-dotenv simplejson
 
+RUN pip install cython
```
jsoules, line 92: For some reason I was getting failures in rebuilding the dev container, that said that scipy couldn't compile because cython wasn't available. This seems sort of contrary to the idea of a dependency manager, but adding an explicit installation of cython seemed to fix it, so... ??

src/extensions/common/hither/useHitherJob.ts
```diff
@@ -50,34 +50,24 @@ const useHitherJob = <T>(functionName: string, functionArgs: {[key: string]: any
...
+    switch (state.status) {
+        case 'finished':
+            retrievedJob.result = state.job?.result
+        // As of right now, for any other case we return the default set above.
+        // Anticipate this might be more customized in the future.
+        // eslint-disable-next-line no-fallthrough
+        case 'running':
+        case 'error':
+        default:
+            break;
```
jsoules, line 68: It turns out we only actually have two cases: either the job is finished (and we return the result), or it is not (and we leave the result undefined). I left it as a switch statement since I figured we might do more customization for specific statuses later, but if that assumption is incorrect we could make this even simpler.

.devcontainer/Dockerfile
magland: Looks good

src/extensions/common/hither/useHitherJob.ts
magland: Looks like a good simplification to me.

magland (approved):

PR #311, opened 2021-01-08
Snippets current timepoint

magland:


magland:
Implemented the following functionality:

* If snippet box is clicked, we set the current timepoint to equal the timepoint for that snippet
* snippet boxes are highlighted if the current timepoint is within 20 timepoints of the current timepoint.

This gives a nice interaction between the time widgets and the snippets view

PR #334, opened 2021-01-14
preload step: download recording/sorting and extract snippets prior to enabling GUI

magland:
The downloading/extracting snippets step can be time consuming, especially for large datasets. Therefore, it is best to make sure these are loaded first prior to allowing the user to interact with the GUI. Here's a screenshot (a modal dialog says "extracting snippets"):

image

### DIFF
--- a/src/extensions/common/hither/HitherJobStatusView.tsx
+++ b/src/extensions/common/hither/HitherJobStatusView.tsx
@@ -5,20 +5,20 @@ import React, { FunctionComponent } from 'react'
 import { HitherJob } from './HitherInterface'
 
 
-const HitherJobStatusView: FunctionComponent<{job?: HitherJob, width?: number, height?: number}> = ({job, width=200, height=200}) => {
+const HitherJobStatusView: FunctionComponent<{job?: HitherJob, message?: string, width?: number, height?: number}> = ({job, message='', width=200, height=200}) => {
     if (!job) return <div>No job</div>
     return (
         <Box display="flex" width={width} height={height}>
             <Box m="auto">
                 {
                     job.status === 'running' ? (
-                        <CircularProgress />
+                        <span>[{message}] <CircularProgress /></span>
                     ): job.status === 'error' ? (
-                        <span>Error: {job.error_message}</span>
+                        <span>Error: {job.error_message} [{message}]</span>
                     ) : job.status === 'pending' ? (
-                        <FontAwesomeIcon icon={faWater} />
+                        <span>[{message}]<FontAwesomeIcon icon={faWater} /></span>
                     ) : (
-                        <span>{job.status}</span>
+                        <span>[{message}] {job.status}</span>
                     )
                 }
             </Box>
--- a/src/extensions/common/hither/createCalculationPool.ts
+++ b/src/extensions/common/hither/createCalculationPool.ts
@@ -1,5 +1,8 @@
 export interface CalculationPool {
     requestSlot: () => Promise<{complete: () => void}>
+    pause: () => void
+    resume: () => void
+    isPaused: () => boolean
 }
 
 const createCalculationPool = ({maxSimultaneous, method}: {maxSimultaneous: number, method?: 'stack' | 'queue'}): CalculationPool => {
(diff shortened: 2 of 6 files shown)


src/extensions/common/hither/HitherJobStatusView.tsx
```diff
@@ -5,20 +5,20 @@ import React, { FunctionComponent } from 'react'
 import { HitherJob } from './HitherInterface'
 
 
-const HitherJobStatusView: FunctionComponent<{job?: HitherJob, width?: number, height?: number}> = ({job, width=200, height=200}) => {
+const HitherJobStatusView: FunctionComponent<{job?: HitherJob, message?: string, width?: number, height?: number}> = ({job, message='', width=200, height=200}) => {
```
magland, line 8: Allow passing a message to HitherJobStatusView (didn't end up using it after all, though)

src/extensions/common/hither/createCalculationPool.ts
```diff
@@ -1,5 +1,8 @@
 export interface CalculationPool {
     requestSlot: () => Promise<{complete: () => void}>
+    pause: () => void
```
magland, line 3: Ability to pause, resume calculation pool. Didn't end up using this, though. But it does seem to work, so maybe in the future.

src/extensions/mountainview/MVSortingView/MVSortingView.tsx
```diff
@@ -76,33 +76,77 @@ export const openViewsReducer: React.Reducer<View[], OpenViewsAction> = (state:
     else return state
 }
 
-const MVSortingView: FunctionComponent<SortingViewProps> = (props) => {
+const area = (a: 'north' | 'south') => {
+    return a
+}
+
+const MVSortingViewWithCheck: FunctionComponent<SortingViewProps> = (props) => {
```
magland, line 83: Create a parent component that does the preloading.
We want this to be slick -- so we DO show the child component (the GUI) but we
(a) prevent interaction
(b) do not send the updated recording/sorting until after it has been computed

The difficult case is when the user selects a new preprocessing (bandpass filter for example), because then we need to start the preprocessing, but the child component should still hold on to the old recording until the preloading is complete. That way the user doesn't see a blank screen.

It took me a while to get this right... but it's working nicely now.

src/extensions/mountainview/MVSortingView/MVSortingView.tsx
```diff
@@ -76,33 +76,77 @@ export const openViewsReducer: React.Reducer<View[], OpenViewsAction> = (state:
...
+    const {recording, sorting} = props
+
+    const [preprocessingSelection, preprocessingSelectionDispatch] = useReducer(preprocessingSelectionReducer, {filterType: 'none'})
```
magland, line 86: The preprocessingSelection/Dispatch is moved to this parent component

src/extensions/mountainview/MVSortingView/MVSortingView.tsx
```diff
@@ -76,33 +76,77 @@ export const openViewsReducer: React.Reducer<View[], OpenViewsAction> = (state:
...
+                }
+            }
+        }
+        else {
+            throw Error(`Unexpected filter type: ${preprocessingSelection.filterType}`)
+        }
+    }, [recording, preprocessingSelection])
+
+    return (
+        <PreloadCheck recording={preprocessedRecording} sorting={sorting}>
```
magland, line 111: The PreloadCheck does the preloading and  passes preloadStatus to the child. The modal dialog is created inside <PreloadCheck />

src/extensions/mountainview/MVSortingView/PreloadCheck.tsx
```diff
@@ -0,0 +1,115 @@
+import { Dialog } from '@material-ui/core';
```
magland, line 1: Here's the new PreloadCheck component -- it does the preloading and updates the props of the child appropriately

src/extensions/mountainview/MVSortingView/PreloadCheck.tsx
```diff
@@ -0,0 +1,115 @@
...
+    const [message, setMessage] = useState<string>('')
+    const runningState = useRef<{sortingObject: any, recordingObject: any}>({sortingObject: sorting.sortingObject, recordingObject: recording.recordingObject})
+
+    const matchesRunningState = useMemo(() => ((x: {recordingObject: any, sortingObject: any}) => (
+        (runningState.current.sortingObject === x.sortingObject) && (runningState.current.recordingObject === x.recordingObject)
+    )), [])
+
+    useEffect(() => {
+        if (status === 'waiting') {
+            runningState.current = {recordingObject, sortingObject}
```
magland, line 33: We need to keep track of the current recording/sorting objects so we can tell if they have been changed. If changed, we abandon the processing.

src/extensions/mountainview/MVSortingView/PreloadCheck.tsx
```diff
@@ -0,0 +1,115 @@
...
+        return React.cloneElement(
+            children,
+            {
+                preloadStatus: status
+            }
+        )
+    }, [children, status])
+
+    // This is important for when the bandpass filter changes so that we don't start calculating things prior to doing the preloading (e.g. snippets extraction)
+    const [lastValidChild, setLastValidChild] = useState<React.ReactElement | null>(null)
```
magland, line 94: This is important for when the bandpass filter changes so that we don't start calculating things prior to doing the preloading (e.g. snippets extraction)

src/extensions/mountainview/mountainview.py
```diff
@@ -1,7 +1,89 @@
 import labbox_ephys as le
+import hither as hi
 
 
 @le.register_sorting_view(name='MVSortingView')
 def MVSortingView(*, sorting: le.LabboxEphysSortingExtractor, recording: le.LabboxEphysRecordingExtractor):
     import labbox_ephys_widgets_jp as lew
     return lew.create_sorting_view('MVSortingView', sorting=sorting, recording=recording)
+
+###################################################################################
```
magland, line 10: Here's the preload processing on the python side.

magland:
@jsoules This is somewhat of a complex PR so I would recommend you try it out in the GUI first so you have a clear idea of what it is trying to do. Be sure to select "bandpass filter" and watch how that works. The key thing is that the views are not updated with the bandpass filter until AFTER the preloading is complete and the modal dialog disappears.

src/extensions/mountainview/MVSortingView/PreloadCheck.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    useEffect(() => {
+        if (status === 'waiting') {
+            runningState.current = {recordingObject, sortingObject}
+            setStatus('running')
+            ;(async () => {
+                try {
+                    setMessage('Checking sorting local...')
+                    const result1 = await hither.createHitherJob('preload_check_sorting_downloaded', {sorting_object: sortingObject}, {useClientCache: false, calculationPool}).wait() as {isLocal: boolean}
+                    if (!matchesRunningState({recordingObject, sortingObject})) return
```
jsoules, line 39: These checks (l. 37 - 71, the whole `try` block for downloading the data) are all very similar--I think it's possible to simplify a lot by introducing a function here. (Would probably make it easier to follow too.)

Also to make sure I'm clear--the purpose of these hither calls is to make sure that the data (respectively, sortings, recordings, and snippets) have been downloaded/processed and are now available locally, right?

Issue #310, opened 2021-01-08
Generalize toolbar widget structure


jsoules:
We don't have a whole lot of these just yet, but items such as [TimeWidgetToolbarNew.tsx](https://github.com/laboratorybox/labbox-ephys/blob/c3633cce161783e13314bdf971e6ca83c19424a6/src/extensions/timeseries/TimeWidgetNew/TimeWidgetToolbarNew.tsx) will soon be joined by other toolbars following a similar structure. The structure already approximates a flexible separation of display and function (with the `customActions` property).

I think it'd be worthwhile to expand that into a toolbar widget framework that handles toolbars generally. Then individual panel views can display a toolbar by passing an array of objects defined in data (as per the custom actions), with an optional 'divider' action type for visual dividers.

Low priority right now.

magland:
I completely agree, and would in fact bump this up to higher priority.

In addition to adding actions, we should make it easy to add electrode selection or unit selection views.

PR #337, opened 2021-01-19
Add average-waveform scaling functions with keypress functionality.

### LINKED ISSUE #268: avg waveform keypress up/down scale amp

jsoules:
Fixes #268.

This PR includes changes that implement a left-panel toolbar with splitter in the `AverageWaveformsView.tsx` collection component which allows button presses or keystrokes to scale the average waveforms displayed. Scaling factor is shared with the time series component, so the two will scale in sync, regardless of which one is controlled.

During the course of this development, I have been thinking about ways that the idea of interactivity can be generalized. Another issue (#310) has been created for this purpose, and I'll have notes to add to that ticket as well--but to give a very brief summary, the use of `customActions` as modeled in the `customActions` prop in [TimeWidgetNew.tsx](https://github.com/laboratorybox/labbox-ephys/blob/0dd6d4269468b2e71d841cbde3d3cbf1c1b8ac61/src/extensions/timeseries/TimeWidgetNew/TimeWidgetNew.tsx#L239) (passed in from [TimeseriesWidgetNew.tsx](https://github.com/laboratorybox/labbox-ephys/blob/0dd6d4269468b2e71d841cbde3d3cbf1c1b8ac61/src/extensions/timeseries/TimeseriesViewNew/TimeseriesWidgetNew.tsx#L226)) points the way to having actions more generally defined at the level of the collection-components, which will hopefully put some of the state logic closer to the reducers and decrease some of the required plumbing. I'll need to explore this idea further.

As a preliminary step, the present PR includes moving the toolbar support out of the `TimeseriesWidgetNew` area into its own first-class file under `common`.

Final note: I've also commented out the `AverageWaveformView.tsx` file, since I'm pretty sure it's deprecated. Consider this a soft removal, to be finalized shortly if nothing actually breaks by removing it.


src/extensions/averagewaveforms/AverageWaveformsView/AverageWaveformsToolbar.tsx
```diff
@@ -0,0 +1,51 @@
+import { IconButton } from '@material-ui/core';
```
jsoules, line 1: Also largely cribbed from `TimeWidgetToolbarNew.tsx`. This can probably be generalized to just be a shared implementation of a toolbar.

src/extensions/averagewaveforms/AverageWaveformsView/AverageWaveformsView.tsx
```diff
@@ -1,34 +1,92 @@
...
+import { ActionItem, DividerItem } from '../../common/Toolbars';
 import { SortingViewProps } from "../../extensionInterface";
+import AverageWaveformsToolbar from './AverageWaveformsToolbar';
 import AverageWaveformView2 from './AverageWaveformView2';
 
-const AverageWaveformsView: FunctionComponent<SortingViewProps> = (props) => {
+export type AverageWaveformAction = ActionItem | DividerItem
+
+
+const TOOLBAR_INITIAL_WIDTH = 55
```
jsoules, line 14: One of the few 'magic numbers' in here; a 55-pixel width happens to look good on my screen. By using a Slider, we make sure that the user can modify it if needed.

Note that the Toolbar is not vertically centered. This looks a little odd (once you notice it) for the average waveform view. However, since they share a scroll bar, there's not really a way to center it vertically. (Presumably solving that layout issue will be higher-yield when we're in a standardized toolbar component regime.)

src/extensions/averagewaveforms/AverageWaveformsView/AverageWaveformsView.tsx
```diff
@@ -1,34 +1,92 @@
...
+
+const AverageWaveformsView: FunctionComponent<SortingViewProps & SizeMeProps> = (props) => {
```
jsoules, line 16: Getting `SizeMe` involved to compute width so we can use the slider. We might not need to do this, but I believe it was used in the time series component set.

src/extensions/averagewaveforms/AverageWaveformsView/AverageWaveformsView.tsx
```diff
@@ -1,34 +1,92 @@
...
+    const width = props.size.width;
+    const height = 650 // hard-coded as per TimeseriesForRecordingView.tsx
 
-    // useCheckForChanges('AverageWaveformsView', props)
+    const _handleScaleAmplitudeUp = useCallback(() => {
+        selectionDispatch({type: 'ScaleAmpScaleFactor', multiplier: 1.15})
+    }, [selectionDispatch])
+    const _handleScaleAmplitudeDown = useCallback(() => {
+        selectionDispatch({type: 'ScaleAmpScaleFactor', multiplier: 1 / 1.15})
+    }, [selectionDispatch])
```
jsoules, line 40: Defines the scaling operation so it can be passed into individual units that have layers (to listen to the keystrokes) and into the toolbar (for button clicks).

src/extensions/averagewaveforms/AverageWaveformsView/AverageWaveformView.tsx
```diff
@@ -1,48 +1,51 @@
-import React, { FunctionComponent } from 'react';
-import { SortingSelection, SortingSelectionDispatch } from '../../extensionInterface';
-import WaveformWidget from './WaveformWidget';
+// Pretty sure this whole component is deprecated, so I'm commenting it out to see if anything breaks.
+export { }; // because linter
```
magland, line 2: Yes, I believe we can safely remove it.

src/extensions/averagewaveforms/AverageWaveformsView/AverageWaveformsView.tsx
```diff
@@ -1,34 +1,92 @@
...
-    return (
-        <SortingUnitPlotGrid
-            sorting={sorting}
-            selection={selection}
-            selectionDispatch={selectionDispatch}
-            unitComponent={unitComponent}
-        />
-    );
+    useEffect(() => {
+        if (scalingActions === null) {
```
jsoules, line 43: As discussed elsewhere, we might want to find a method to do this that isn't faking a singleton approach. I'm not sure under what circumstances the action set would need to be updated...

src/extensions/averagewaveforms/AverageWaveformsView/waveformLayer.ts
```diff
@@ -17,6 +18,24 @@ const initialLayerState = {
     electrodeBoxes: []
 }
 
+type LocalLayerProps = LayerProps & { customActions: (ActionItem | DividerItem)[] }
+
+// If any custom actions have been set (that is, something a user of this component wants to happen in response to a key press)
+// expect them to have been passed in with the key 'customActions' & call them here.
+export const handleKeyboardEvent: KeyboardEventHandler = (e: KeyboardEvent, layer: CanvasWidgetLayer<LocalLayerProps, LayerState>): boolean => {
```
jsoules, line 25: It's possible that processing custom actions could be assumed by default--so this could be part of the `CanvasWidgetLayer` rather than manually added to specific layers. (Not sure if that would result in double-processing some actions or in needless empty checks, though.)

src/extensions/averagewaveforms/AverageWaveformsView/AverageWaveformsView.tsx
magland: I confirmed that this can be replaced by the following:

```
useEffect(() => {
        const actions: AverageWaveformAction[] = [
            {
                type: 'button',
                callback: _handleScaleAmplitudeUp,
                title: 'Scale amplitude up [up arrow]',
                icon: <FaArrowUp />,
                keyCode: 38
            },
            {
                type: 'button',
                callback: _handleScaleAmplitudeDown,
                title: 'Scale amplitude down [down arrow]',
                icon: <FaArrowDown />,
                keyCode: 40
            }
        ]
        setScalingActions(actions)
    }, [_handleScaleAmplitudeUp, _handleScaleAmplitudeDown])
```

The scaling actions only depend on _handleScaleAmplitudeUp/Down. Therefore they only need to be set when those change.

src/extensions/averagewaveforms/AverageWaveformsView/AverageWaveformsToolbar.tsx
magland: Yes, it doesn't really have anything to do with average waveforms. So we can pull this out to something like ViewToolbar in common... and pass in the custom actions.

src/extensions/averagewaveforms/AverageWaveformsView/waveformLayer.ts
magland: we'd need to be careful... i don't think it's a big problem to manually add to specific layers.

magland (approved):
i had one comment with code simplification. other than that, i confirm it works, and it is ready for merge.

src/extensions/averagewaveforms/AverageWaveformsView/AverageWaveformView.tsx
jsoules: Removed.

src/extensions/averagewaveforms/AverageWaveformsView/AverageWaveformsView.tsx
jsoules: Replaced (i.e. removed the attempted set-once check and dependency on the `actions` and `setScalingActions` properties). I did a little checking to confirm that it still seems to load only once.

jsoules:
Removed old `AverageWaveformView` and renamed `AverageWaveformView2` to use that name.

PR #342, opened 2021-01-20
decorator for serializing np hither outputs

magland:


magland:
Added a @le.serialize decorator that converts numpy outputs of hither functions to json-able objects (with the numpy data imbedded as base64 strings). These are the deserialized on the client (javascript) side. This used to be done as part of the labbox framework, but the advantage of doing this with the decorator is that when a remote compute resource is used, the result is serialized and cached on the remote computer in exactly the same form as when it is consumed on the javascript client side. This minimizes the number of serialization/deserialization steps when using a remote compute resource.

jsoules (approved):
Looks fine to me

PR #345, opened 2021-01-21
restore job monitor functionality

magland:


src/AppContainer.tsx
```diff
@@ -4,20 +4,31 @@ import Toolbar from '@material-ui/core/Toolbar';
 // LABBOX-CUSTOM /////////////////////////////////////////////
 import Typography from '@material-ui/core/Typography';
```
magland, line 5: Converted this file to .tsx

src/components/HitherJobMonitor/HitherJobMonitor.tsx
```diff
@@ -1,28 +1,38 @@
...
-}
-
-interface DispatchProps {
-}
+type Props = {}
 
-interface OwnProps {
-}
+const HitherJobMonitor: FunctionComponent<Props> = () => {
+    const [hitherJobs, setHitherJobs] = useState<HitherJob[]>([])
```
magland, line 10: The hitherJobs is no longer part of the redux state. Instead we get it from the hitherInterface via useContext

src/components/HitherJobMonitor/HitherJobMonitor.tsx
```diff
@@ -1,28 +1,38 @@
...
+    const hither = useContext(HitherContext)
+    useEffect(() => {
+        // this should only get called once
+        // (hither should not change, but if it does we might have a problem here)
+        const update = () => {
+            const hj = hither.getHitherJobs()
+            setHitherJobs(hj)
```
magland, line 17: We do polling to get the updated hither jobs (which is sufficient for monitory changes). This way we don't have an expensive state changes whenever a job status changes. This useEffect with cleanup is a nice mechanism for doing polling of this nature.

src/components/HitherJobMonitor/HitherJobMonitorControl.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { getPathQuery } from '../../kachery';
+import { DocumentInfo } from '../../reducers/documentInfo';
+
+type Props = {
+    documentInfo: DocumentInfo,
+    websocketStatus: string
+}
+
+const HitherJobMonitorControl: FunctionComponent<Props> = ({ documentInfo, websocketStatus }) => {
+    const [hitherJobs, setHitherJobs] = useState<HitherJob[]>([])
```
magland, line 13: Same technique as in HitherJobMonitor

src/extensions/common/hither/HitherInterface.ts
```diff
@@ -42,8 +42,10 @@ export const dummyHitherJob: HitherJob = {
 
 export interface HitherInterface {
     createHitherJob: (functionName: string, kwargs: {[key: string]: any}, opts: HitherJobOpts) => HitherJob
+    getHitherJobs: () => HitherJob[]
```
magland, line 45: Access hither jobs through the hither interface context

src/components/HitherJobMonitor/HitherJobMonitorControl.tsx
jsoules: Ah yes. Wondering if the function could be extracted. Would it be possible to make it part of the `hither` context object? In addition to avoiding the repeat code, that maybe could help mitigate any issues associated with the `hither` object somehow disappearing under the feet of this effect hook?

PR #350, opened 2021-01-21
Finish merge of PRs 342 and 345.

jsoules:


jsoules:
Merged the `compute-resource` branch PRs back into `compute-resource`, despite its having been merged already; and re-merged current master (which includes the previous `compute-resource` changes) back in. A subsequent squash-merge should fix this up. It seemed the easiest way, and I think this puts everything in line.

However, while I think everything matches the intended final state, I'm noticing both some errors in the linter (incorrect loading of the `kachery` and `labbox_ephys` python packages) and in the interface (when I run this version on http://localhost:15351/default/sorting/XyVPAErc8c the time series doesn't seem to load correctly). I'm not sure exactly what's happening, but as it looks like this might have somewhere introduced some breaking changes, I don't want to merge it in until that's been ironed out.

So we'll need at least a little more proofing before this is ready.

magland:
Seems to work on my end. I will merge into master and we'll see if any problems arise.

PR #363, opened 2021-01-27
Import recordings: snippet for importing from spikeforest

magland:


python/labbox_ephys/import_to_workspace.py
```diff
@@ -75,32 +75,39 @@ def _import_le_sorting(sortings_subfeed: kp.Subfeed, le_sorting):
         }
     })
 
-def import_recording(*, feed: kp.Feed, workspace_name: str, recording: se.RecordingExtractor, recording_id: str):
+def random_id():
+    return str(uuid.uuid4())[-12:]
+
+def import_recording(*, feed: kp.Feed, workspace_name: str, recording: se.RecordingExtractor, recording_label: str):
+    recording_id = 'R-' + random_id()
```
magland, line 82: Now, when importing a recording, the recording ID is randomly generated. The recording label is what displays on the GUI.

python/labbox_ephys/import_to_workspace.py
```diff
@@ -75,32 +75,39 @@ def _import_le_sorting(sortings_subfeed: kp.Subfeed, le_sorting):
...
-        'description': f'Imported from Python: {recording_id}'
+        'description': f'Imported from Python: {recording_label}'
     }
     recordings_subfeed = feed.get_subfeed(dict(workspaceName=workspace_name, key='recordings'))
     _import_le_recording(recordings_subfeed, x)
+    return x
 
-def import_sorting(*, feed: kp.Feed, workspace_name: str, recording: se.RecordingExtractor, sorting: se.SortingExtractor, recording_id: str, sorting_id: str):
+def import_sorting(*, feed: kp.Feed, workspace_name: str, recording: se.RecordingExtractor, sorting: se.SortingExtractor, recording_id: str, sorting_label: str):
+    sorting_id = 'S-' + random_id()
```
magland, line 95: Similar for importing a sorting... the sorting ID is randomly generated (the sorting_label is what displays in the GUI)

src/containers/ImportRecordingsInstructions.tsx
```diff
@@ -17,6 +18,10 @@ const ImportRecordingsInstructions: FunctionComponent<Props> = () => {
                 <CopyToClipboardButton text={import_example_simulated_recording_py} />
                 <Markdown source={mdWrapPy(import_example_simulated_recording_py)} />
             </Expandable>
+            <Expandable label="Import SpikeForest recordings">
```
magland, line 21: Now there are two example scripts for importing recordings. (a) simulation or (b) from spikeforest

src/containers/importExamples/import_example_simulated_recording.py
```diff
@@ -20,8 +20,8 @@ def prepare_recording_sorting():
     return R, S
 
 recording, sorting_true = prepare_recording_sorting()
-sorting_id = recording_id + ':true'
+sorting_label = 'true'
```
magland, line 23: The label of the sorting can just be 'true' (does not have to contain the recording id as before) because now the sorting ID is randomly generated - and the label does not need to be unique across all sortings in the workspace.

src/containers/importExamples/import_spikeforest_recording.py
```diff
@@ -0,0 +1,41 @@
+import spikeextractors as se
```
magland, line 1: This is the Python snippet for importing from spikeforest

src/containers/importExamples/import_spikeforest_recording.py
```diff
@@ -0,0 +1,41 @@
...
+    "recording_uri": "sha1://759c81e6e20d3c352eaff6ce5d4dc7f948c45f25/SF/PAIRED_BOYDEN/paired_boyden32c/419_1_7.json",
+    "sorting_true_uri": "sha1://a5b7de64d7a2a96c0e1af70cdb2d4ae927e4949d/firings_true.mda"
+}
+X3 = {
+    "label": "SF/PAIRED_KAMPFF/paired_kampff/2014_11_25_Pair_3_0",
+    "recording_uri": "sha1://a205f87cef8b7f86df7a09cddbc79a1fbe5df60f/SF/PAIRED_KAMPFF/paired_kampff/2014_11_25_Pair_3_0.json",
+    "sorting_true_uri": "sha1://1cd517687aeca7ecdfaa9695680038d142a75031/firings_true.mda"
+}
+# To find more examples, see: https://github.com/flatironinstitute/spikeforest_recordings
+# However: note that some processing needs to be done to the files in this repo (to add the manifests to the raw data). This is WIP
```
magland, line 23: This is an important comment... we need to process the spikeforest_recordings repo so that the uri's of raw files have manifests.

python/labbox_ephys/import_to_workspace.py
```diff
@@ -75,32 +75,39 @@ def _import_le_sorting(sortings_subfeed: kp.Subfeed, le_sorting):
         }
     })
 
-def import_recording(*, feed: kp.Feed, workspace_name: str, recording: se.RecordingExtractor, recording_id: str):
+def random_id():
+    return str(uuid.uuid4())[-12:]
```
jsoules, line 79: Any reason to limit it to 12 characters? (I suppose that's still an awful lot of possible random numbers...)

src/containers/importExamples/import_spikeforest_recording.py
jsoules: Ok. What would need to be done to generate manifests for those?

src/containers/importExamples/import_example_simulated_recording.py
jsoules: That's great! Will we have a facility to change the label of a sorting later on through the interface, or does it have to be set to the final value at import time?

jsoules (approved):
This one seems pretty straightforward.

python/labbox_ephys/import_to_workspace.py
magland: the structure of uuid4 is there is a group of 12 characters at the end. There won't be any penalty for changing this later if we need to. I don't know much about uuid4, so it would be good to know if this is a reasonable thing to do.

magland:
> That's great! Will we have a facility to change the label of a sorting later on through the interface, or does it have to be set to the final value at import time?

The plan is to be able to change the label later

magland:
> Ok. What would need to be done to generate manifests for those?

We should chat about that today.

python/labbox_ephys/import_to_workspace.py
jsoules: I spent some time this morning reading up on it. It looks like the conclusion is that depending on implementation, the tail end of the UUID is likely to be the 'most random', but obviously decreasing the overall bit count increases the chances of getting a collision (even if that chance is pretty low for a 12-digit number).

That said, these IDs aren't really meant to be human-readable, so I'm not sure there's much benefit to limiting it (unless we were trying to avoid having non-numeric characters)?

magland: The reason to limit it is that these go in the URL for labbox-ephys. sqrt(16^12) is around 16 million. And these are specific to the given workspace, so I think we're pretty safe. Worst that can happen (very rare) is you get an error on import, and then just try again.

PR #367, opened 2021-01-28
import nwb recording/sorting

magland:


python/labbox_ephys/extractors/bandpass_filter.py
```diff
@@ -1,11 +1,24 @@
+import spikeextractors as se
```
magland, line 1: The bandpass filter (preprocessing) was inefficient, so I made some adjustments.

python/labbox_ephys/extractors/bandpass_filter.py
```diff
@@ -1,11 +1,24 @@
...
+import scipy.fft
+import math
 
 
 class BandpassFilterRecording(FilterRecording):
-    def __init__(self, *, recording, freq_min, freq_max, freq_wid):
-        FilterRecording.__init__(self, recording=recording, chunk_size=3000 * 10)
+    def __init__(self, *, recording: se.RecordingExtractor, freq_min, freq_max, freq_wid):
+        self._padding = 3000
+        target_ram = 100 * 1000 * 1000
```
magland, line 12: Target number of bytes of RAM in a chunk

python/labbox_ephys/extractors/bandpass_filter.py
```diff
@@ -1,11 +1,24 @@
...
+        target_chunk_size = math.ceil(
+            min(
+                recording.get_sampling_frequency() * 30,
+                target_ram / (recording.get_num_channels() * 4)
+            )
+        )
+        # It's important that the fft's have size 2^x. So we prepare the chunk sizes to have size 2^x - 2*padding
+        chunk_size = int(2 ** math.ceil(np.log2(target_chunk_size)) - self._padding * 2)
```
magland, line 20: FFT is most efficient for chunks of size that is a power of 2

python/labbox_ephys/extractors/bandpass_filter.py
```diff
@@ -63,15 +75,15 @@ def _do_filter(self, chunk):
         #    for m in range(M):
         #        chunk2[m,:]=chunk2[m,:]-np.mean(chunk2[m,:])
         # Do the actual filtering with a DFT with real input
-        chunk_fft = np.fft.rfft(chunk2)
+        chunk_fft = scipy.fft.rfft(chunk2, workers=1)
```
magland, line 78: Switched from np.fft.rfft to scipy.fft.rfft. Probably doesn't make too much of a difference. But with scipy I could specify the number of workers = 1. That way we don't have as much of a chance of overloading the CPUs if many tasks are happening simultaneously.

python/labbox_ephys/extractors/labboxephysrecordingextractor.py
```diff
@@ -222,9 +220,10 @@ def __init__(self, arg: Union[str, dict], download: bool=False):
             self._recording: se.RecordingExtractor = MdaRecordingExtractor(timeseries_path=data['raw'], samplerate=data['params']['samplerate'], geom=np.array(data['geom']), download=download)
         elif recording_format == 'nrs':
             self._recording: se.RecordingExtractor = NrsRecordingExtractor(**data)
-        # elif recording_format == 'nwb':
-        #     path0 = kp.load_file(data['path'])
-        #     self._recording: se.RecordingExtractor = NwbRecordingExtractor(path0, electrical_series_name='e-series')
+        elif recording_format == 'nwb':
+            from .nwbextractors import NwbRecordingExtractor
```
magland, line 224: Here's where we support the 'nwb' format. Note that we use a custom NwbRecordingExtractor because the spikeinterface one has a couple of issues.

python/labbox_ephys/extractors/labboxephyssortingextractor.py
```diff
@@ -106,6 +106,10 @@ def __init__(self, arg, samplerate=None):
             S.set_sampling_frequency(samplerate)
             S.set_times_labels(times_npy.ravel(), labels_npy.ravel())
             self._sorting = S
+        elif sorting_format == 'nwb':
+            from .nwbextractors import NwbSortingExtractor
```
magland, line 110: Similar for NwbSortingExtractor

src/containers/ImportRecordingsInstructions.tsx
```diff
@@ -22,6 +23,10 @@ const ImportRecordingsInstructions: FunctionComponent<Props> = () => {
                 <CopyToClipboardButton text={import_spikeforest_recording_py} />
                 <Markdown source={mdWrapPy(import_spikeforest_recording_py)} />
             </Expandable>
+            <Expandable label="Import NWB recordings">
```
magland, line 26: Here's where we provide the instructions (example python snippet) for importing nwb recording/sorting

src/containers/importExamples/import_nwb_recording.py
```diff
@@ -0,0 +1,36 @@
+import spikeextractors as se
```
magland, line 1: Here's the example script for importing from nwb formatted files

src/extensions/clusters/clusters.py
```diff
@@ -28,11 +29,17 @@ def createjob_individual_cluster_features(labbox, recording_object, sorting_obje
...
+        return range(n)
+    incr = n / m
+    return [int(math.floor(i * incr)) for i in range(m)]
+
+@hi.function('individual_cluster_features', '0.1.3')
 @hi.container('docker://magland/labbox-ephys-processing:0.3.19')
 @hi.local_modules([os.getenv('LABBOX_EPHYS_PYTHON_MODULE_DIR')])
 @le.serialize
-def individual_cluster_features(snippets_h5, unit_id):
+def individual_cluster_features(snippets_h5, unit_id, max_num_events=1000):
```
magland, line 42: Limit number of events to a maximum of 1000 for cluster features. Eventually we will want this to be a parameter in the GUI

src/extensions/mountainview/mountainview.py
```diff
@@ -116,5 +116,5 @@ def preload_extract_snippets(labbox, recording_object, sorting_object):
         job_handler=jh,
         container=jh.is_remote
     ):
-        snippets_h5 = prepare_snippets_h5.run(recording_object=recording_object, sorting_object=sorting_object, max_events_per_unit=None)
+        snippets_h5 = prepare_snippets_h5.run(recording_object=recording_object, sorting_object=sorting_object)
```
magland, line 119: Uniformize how we call prepare_snippets_h5

src/extensions/spikeamplitudes/SpikeAmplitudesView/SpikeAmplitudesPanel.ts
```diff
@@ -23,7 +23,7 @@ class SpikeAmplitudesPanel {
     _calculationScheduled: boolean = false
     _calculationError: Error | null = null
     _yrange: {min: number, max: number} | null = null
-    _panelGroup: SpikeAmplitudesPanel[] | null = null
+    _globalAmplitudeRange: {min: number, max: number} | null = null
```
magland, line 26: Changed method for determining the auto-amplitude range for the spike amplitude plot. There was a problem (react call stack overflow) with the old method, and this is more straightforward anyway

src/extensions/timeseries/TimeseriesViewNew/useTimeseriesModel.ts
```diff
@@ -3,8 +3,8 @@ import { createCalculationPool, HitherContext, HitherInterface } from "../../com
 import useFetchCache from "../../common/useFetchCache"
 import { RecordingInfo } from "../../extensionInterface"
 
-// it may be important to limit this to 1 at a time when using a filter
-const timeseriesCalculationPool = createCalculationPool({maxSimultaneous: 1, method: 'stack'})
+// it may be important to limit this when using a filter
+const timeseriesCalculationPool = createCalculationPool({maxSimultaneous: 4, method: 'stack'})
```
magland, line 7: Increase maxSimultaneous here from 1 to 4... because now we know that the number of worker threads is limited to 1 for each bandpass filter process.

src/extensions/spikeamplitudes/SpikeAmplitudesView/SpikeAmplitudesTimeWidget.tsx
```diff
@@ -29,17 +29,26 @@ const SpikeAmplitudesTimeWidget: FunctionComponent<Props> = ({ spikeAmplitudesDa
 
     useEffect(() => {
         const panels: SpikeAmplitudesPanel[] = []
+        const allMins: number[] = []
+        const allMaxs: number[] = []
+
         unitIds.forEach(unitId => {
             const p = new SpikeAmplitudesPanel({spikeAmplitudesData, recording, sorting, unitId, hither})
+            const amps = spikeAmplitudesData.getSpikeAmplitudes(unitId)
```
magland, line 37: This stuff all goes along with new method for determining the auto scale factor for spike amplitudes view

python/labbox_ephys/extractors/bandpass_filter.py
magland: Yeah... I don't really know how to set it.

To consider: FFT scales as n*log(n), so it's not quite linear in n. Therefore there is a penalty for larger chunks. On the other hand there are other reasons why too small chunks are less efficient. So there is a sweet spot, and this was my best attempt. This can be improved, but not sure how much we will gain.

python/labbox_ephys/extractors/bandpass_filter.py
magland: There are "edge effects" in the bandpass filter. So the filter is not accurate at the edges of the chunk. Therefore we need some padding, and overlap.

magland:
@jsoules, regarding nwbextractors.py
I grabbed that from the current spikeinterface. However, there are some issues with that code, so I needed to adjust in a couple of places. We should be responsible and push our modifications back spikeextractors.
However, I like having the .nwbextractors code embedded in this repo so we don't risk it breaking over time. Will need to think about how to best synchronize.

Issue #376, opened 2021-01-29
Put getMin and getMax in the common/ directory


magland:
These two functions have shown up in several different files now, it'd be a good idea to put them in a library

_Originally posted by @jsoules in https://github.com/laboratorybox/labbox-ephys/pull/375#discussion_r566883806_

PR #378, opened 2021-01-29
(Fix Issue #376) Implement getMin, getMax as utility functions in existing library; fix compile warnings.

jsoules:
Created these two functions in `utils/Utility.ts`, a library file that apparently has existed for some time (who knew?)

Text of those functions is literally just copied over, though I gave them & their internal variables ever-so-slightly more descriptive names.

Instances of locally-defined `getMax` and `getMin` have been replaced to use the library functions.

I also corrected some missing or superfluous dependencies in `src/extensions/timeseries/TimeseriesViewNew/useTimeseriesModel.ts` and `src/extensions/timeseries/TimeWidgetNew/TimeWidgetNew.tsx`, and added a few 'ignore this' comments to the linter for one of those files (in re: non-referenced variables that I believe are there for a good reason).


magland:
Looks good, except this needs to go somewhere inside src/extensions. That way the functions can be shared between web gui and jupyter code.

magland:
I was trying to obey the rule that anything inside extensions/ does not refer to anything above extensions/. That we it is an independent folder that can be placed in various contexts. Otherwise we need to line it up properly with respect to other source.

jsoules:
Hadn't seen your comment--once I realized it wasn't compiling, I dug around and figured out that it was because of the changed library path, so I added `util` to codesync and gitignore.

However, while it now compiles locally, I'm still getting the check failing on the remote. I'd really rather keep it in its own directory since it isn't properly part of the extension system, but if you think it makes more sense I can just drop it in `extensions/common/utililty.ts` or something.

jsoules:
Okay, will move to extensions/common.

magland:
There are a lot of things inside extensions/ that don't really belong there. Haven't solved that issue yet. I think we should put util in there fore now.

PR #391, opened 2021-02-04
fix rotateRect()

### LINKED ISSUE #377: fix rotateRect()
I'm not sure this is right: if we take a rectangular region from (1, 1) to (3, 3), then xmin = 1, xmax = 3, ymin = 1, ymax = 3.
But by this transformation we would get a new region (-3, 3) to (-1, 1): new_ymax < new_ymin. Doesn't this result in a twist or reflection?

_Originally posted by @jsoules in https://github.com/laboratorybox/labbox-ephys/pull/375#discussion_r566869406_

magland:
Fixes #377

### DIFF
--- a/src/extensions/common/CanvasWidget/CanvasPainter.ts
+++ b/src/extensions/common/CanvasWidget/CanvasPainter.ts
@@ -34,7 +34,7 @@ interface TextAlignmentConfig {
 }
 
 const rotateRect = (r: RectangularRegion) => {
-    return {xmin: -r.ymax, xmax: -r.ymin, ymin: r.xmax, ymax: r.xmin}
+    return {xmin: -r.ymax, xmax: -r.ymin, ymin: r.xmin, ymax: r.xmax}
 }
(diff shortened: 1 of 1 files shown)


src/extensions/common/CanvasWidget/CanvasPainter.ts
```diff
@@ -34,7 +34,7 @@ interface TextAlignmentConfig {
 }
 
 const rotateRect = (r: RectangularRegion) => {
-    return {xmin: -r.ymax, xmax: -r.ymin, ymin: r.xmax, ymax: r.xmin}
+    return {xmin: -r.ymax, xmax: -r.ymin, ymin: r.xmin, ymax: r.xmax}
```
jsoules, line 37: Okay. So this corresponds to a rotation of 90 degrees (counterclockwise) around the origin.

I did a few sketches and it looks to check out.

jsoules (approved):

PR #394, opened 2021-02-04
fix ts compile error

magland:
Needed to be more explicit with typing. Typescript was complaining in the github action/workflow even though it wasn't complaining locally.


magland:
Merging this now: just a simple ts compile error fix for jupyter.

PR #401, opened 2021-02-04
useTimeseriesModel: comment code for retrieving segments

### LINKED ISSUE #306: useTimeseriesModel: comment code for retrieving segments
I understand this is how to fetch different segments based on where we are in the time series (yes?) but the subtleties of the mechanics are escaping me right now. It's also not immediately clear what `src` and `dst` are referring to. I think some comments would be helpful here.

_Originally posted by @jsoules in https://github.com/laboratorybox/labbox-ephys/pull/301#discussion_r554056798_

magland:
Fixes #306


jsoules (approved):
This is great. Very helpful.

Issue #385, opened 2021-02-02
support SpikeInterface pickle format


magland:
Will work with Alessio to figure out how best to do this.

magland:
@alejoe91, Here's what I am concerned about with regards to supporting the SpikeInterface pkl format as is.

Let's consider the following for reference:
```python
{'class': 'spiketoolkit.preprocessing.bandpass_filter.BandpassFilterRecording',
 'module': 'spiketoolkit',
 'kwargs': {'recording': {'class': 'spikeextractors.subrecordingextractor.SubRecordingExtractor',
   'module': 'spikeextractors',
   'kwargs': {'parent_recording': {'class': 'spikeextractors.extractors.mdaextractors.mdaextractors.MdaRecordingExtractor',
     'module': 'spikeextractors',
     'kwargs': {'folder_path': '/home/magland/src/labbox-ephys/devel/example_notebooks/test'},
     'key_properties': {'group': array([0, 0, 0, 0]),
      'location': array([[1.00000000e+000, 0.00000000e+000, 4.67369358e-310],
             [2.00000000e+000, 0.00000000e+000, 5.58294180e-322],
             [3.00000000e+000, 0.00000000e+000, 1.58101007e-322],
             [4.00000000e+000, 0.00000000e+000, 6.91887545e-310]])},
     'version': '0.9.3',
     'dumpable': True},
    'channel_ids': None,
    'renamed_channel_ids': None,
    'start_frame': 10,
    'end_frame': None},
   'key_properties': {'group': array([0, 0, 0, 0]),
    'location': array([[1.00000000e+000, 0.00000000e+000, 4.67369358e-310],
           [2.00000000e+000, 0.00000000e+000, 5.58294180e-322],
           [3.00000000e+000, 0.00000000e+000, 1.58101007e-322],
           [4.00000000e+000, 0.00000000e+000, 6.91887545e-310]])},
   'version': '0.9.3',
   'dumpable': True},
  'freq_min': 300,
  'freq_max': 6000,
  'freq_wid': 1000,
  'filter_type': 'fft',
  'order': 3,
  'chunk_size': 30000,
  'cache_chunks': False},
 'key_properties': {'group': array([0, 0, 0, 0]),
  'location': array([[1.00000000e+000, 0.00000000e+000, 4.67369358e-310],
         [2.00000000e+000, 0.00000000e+000, 5.58294180e-322],
[...]

PR #402, opened 2021-02-04
spikeinterface recording serialization -> labbox

magland:


magland:
@alejoe91

Here's a prelim example of converting spikeinterface dump-to-dict into labbox dict.

See: https://github.com/flatironinstitute/labbox-ephys/blob/magland/issue385/devel/example_spikeinterface_to_labbox.py

magland:
Relates to #385

PR #403, opened 2021-02-05
Subfeed messages: improve client/server msg passing 

magland:


jupyterlab/labbox_ephys_widgets_jp/labbox_ephys_widgets_jp/create_recording_view.py
```diff
@@ -48,9 +48,9 @@ def __init__(self) -> None:
             super().__init__()
             self.on_msg(self._handle_message)
             self._worker_session = le.WorkerSession(labbox_config=labbox_config)
-            def on_msg(msg):
-                self.send(msg)
-            self._worker_session.on_message(on_msg)
+            def on_msgs(msgs):
```
magland, line 51: Send lists of messages rather than individual messages from server->client. This should improve efficiency (avoid unnecessary interim re-renders on the client), but also allows sending groups of messages that must be processed together on the client.

jupyterlab/labbox_ephys_widgets_jp/labbox_ephys_widgets_jp/create_sorting_view.py
```diff
@@ -93,8 +93,8 @@ def _handle_message(self, widget, msg, buffers):
             elif msg['type'] == 'addSubfeedWatch':
                 self._worker_session.add_subfeed_watch(
                     watch_name=msg['watchName'],
-                    feed_id=msg['feedId'],
-                    subfeed_hash=msg['subfeedHash']
+                    feed_uri=msg['feedUri'],
```
magland, line 96: Pass feed_uri and subfeed_name rather than feed_id and subfeed_hash.

Note that feed_uri and subfeed_hash can always be derived from feed_id and subfeed_name.

jupyterlab/labbox_ephys_widgets_jp/src/pluginWidgets.tsx
```diff
@@ -364,24 +364,28 @@ export class SortingView extends DOMWidgetView {
...
-      }
-      else if (msg.type === 'hitherJobError') {
-        _startIterating(300)
-        hither.handleHitherJobError(msg)
-      }
-      else if (msg.type === 'debug') {
-        console.info('DEBUG MESSAGE', msg)
+    this.model.on('msg:custom', (msgs: any) => {
+      console.info('RECEIVED MESSAGES', msgs)
+      for (let msg of msgs) {
```
magland, line 373: Handle list of messages rather than individual messages as described above.

python/labbox_ephys/api/_workersession.py
```diff
@@ -71,28 +73,21 @@ def handle_message(self, msg):
             self._readonly = msg['clientInfo']['readOnly']
             if not self._feed_uri:
                 self._feed_uri = 'feed://' + self._default_feed_id
-                # self._feed_uri = kp.create_feed(feed_name='labbox-ephys-default').get_uri()
-            # assert self._feed_uri.startswith('sha1://'), 'For now, feedUri must start with sha1://'
             self._feed = kp.load_feed(self._feed_uri)
-            for key in ['recordings', 'sortings']:
```
magland, line 77: Now recordings and sortings subfeeds are not treated as special watches... the client now has control for adding arbitrary subfeed watches (rather than the server). This puts us in a position of implementing workspace view in jupyter.

src/index.tsx
```diff
@@ -217,6 +230,22 @@ hither._registerSendMessage((msg) => {
 })
 
 const handleSetWorkspaceInfo = (workspaceInfo: WorkspaceInfo) => {
+  // important to send the subfeed watch messages before the reportClientInfo
+  // because otherwise our initial load will not include the recordings/sortings
+  apiConnection.sendMessage({
+    type: 'addSubfeedWatch',
```
magland, line 236: Here is where we make the subfeed watches for recordings/sortings in the workspace. Before this was done on the server side. Now it puts us in a position to implement workspace view in jupyter.

python/labbox_ephys/api/_workersession.py
```diff
@@ -273,3 +271,24 @@ def _is_jsonable(x) -> bool:
...
         return False
+
+def _feed_id_from_uri(uri: str):
+    a = uri.split('/')
+    return a[2]
+
+def _subfeed_hash_from_name(subfeed_name: Union[str, dict]):
+    if isinstance(subfeed_name, str):
+        if subfeed_name.startswith('~'):
+            return subfeed_name[1:]
```
jsoules, line 282: To be clear, any name starting with a tilde is assumed to be actually a hash. Is there a capacity for the user to pick their own name for a subfeed? If so, we should probably warn them not to use this symbol at the beginning of the feed id (which someone might be tempted to do, since it's common as a shorthand for one's home directory).

magland: yes, I created an issue for this.

python/labbox_ephys/api/_workersession.py
```diff
@@ -273,3 +271,24 @@ def _is_jsonable(x) -> bool:
         return True
     except:
         return False
+
+def _feed_id_from_uri(uri: str):
+    a = uri.split('/')
```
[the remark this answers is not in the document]
magland: Created an issue

PR #405, opened 2021-02-05
Move WorkspaceView to src/extensions/

magland:
This PR is mainly about moving files around and updating what props get passed to who. No functionality changed. Preparing to implement WorkspaceView in jupyter - so the WorkspaceView files need to get moved into extensions/ where the files are shared between web gui and jupyter.


magland:
The following PR should be reviewed and merged first:
https://github.com/flatironinstitute/labbox-ephys/pull/403

jupyterlab/labbox_ephys_widgets_jp/labbox_ephys_widgets_jp/create_recording_view.py
```diff
@@ -48,9 +48,9 @@ def __init__(self) -> None:
             super().__init__()
             self.on_msg(self._handle_message)
             self._worker_session = le.WorkerSession(labbox_config=labbox_config)
-            def on_msg(msg):
-                self.send(msg)
-            self._worker_session.on_message(on_msg)
+            def on_msgs(msgs):
```
magland, line 51: Just renaming msg -> msgs

jupyterlab/labbox_ephys_widgets_jp/labbox_ephys_widgets_jp/create_sorting_view.py
```diff
@@ -93,8 +93,8 @@ def _handle_message(self, widget, msg, buffers):
             elif msg['type'] == 'addSubfeedWatch':
                 self._worker_session.add_subfeed_watch(
                     watch_name=msg['watchName'],
-                    feed_id=msg['feedId'],
-                    subfeed_hash=msg['subfeedHash']
+                    feed_uri=msg['feedUri'],
```
magland, line 96: When communicating between server/client we are going to use feedUri and subfeedName rather than feedId and subfeedHash. That's because the latter can be derived from the former, but not the other way around.

magland:
In this PR we are preparing to implement WorkspaceView in jupyter (as WorkspaceViewJp). To do that, we need to move the WorkspaceView code into src/extensions so that it can be shared between web gui and jupyter. Anything inside extensions/ cannot depend on anything in parent directories. So other things need to get moved around as well and we are pulling away from the redux model.

src/components/HitherJobMonitor/HitherJobMonitor.tsx
```diff
@@ -1,8 +1,8 @@
 import { Button, IconButton, Link as LinkMui, Table, TableBody, TableCell, TableHead, TableRow } from '@material-ui/core';
 import { Delete } from '@material-ui/icons';
 import React, { FunctionComponent, useContext, useEffect, useState } from 'react';
-import NiceTable from '../../components/NiceTable';
 import { HitherContext, HitherJob } from '../../extensions/common/hither';
+import NiceTable from '../../extensions/common/NiceTable';
```
magland, line 5: Just relocating files

src/containers/Home.tsx
```diff
@@ -19,6 +19,7 @@ interface StateProps {
 
 interface DispatchProps {
   onDeleteRecordings: (recordingIds: string[]) => void
+  onDeleteSortings: (sortingIds: string[]) => void
```
magland, line 22: This now needs to be passed in explicitly into WorkspaceView, because WorkspaceView no longer has access to redux dispatch.

src/extensions/WorkspaceView/SortingsView.tsx
```diff
@@ -0,0 +1,31 @@
+import { Button, Paper } from "@material-ui/core";
```
magland, line 1: This was just moved / tidied.

src/extensions/WorkspaceView/WorkspaceView.tsx
```diff
@@ -1,21 +1,35 @@
...
-import { Plugins } from '../extensions/extensionInterface';
-import { Recording } from '../reducers/recordings';
-import { ServerInfo } from '../reducers/serverInfo';
-import { Sorting } from '../reducers/sortings';
+import { ExternalSortingUnitMetric, Plugins, Recording, Sorting, SortingCurationAction } from '../extensionInterface';
 import SortingView from './SortingView';
 import WorkspaceRecordingsView from './WorkspaceRecordingsView';
 import WorkspaceRecordingView from './WorkspaceRecordingView';
 
+interface Location {
```
magland, line 7: Making interfaces for Location and History so that WorkspaceView does not depend on react-router-dom

src/extensions/common/NiceTable.tsx
```diff
@@ -3,9 +3,10 @@ import { Delete, Edit } from "@material-ui/icons";
 import React, { FunctionComponent } from 'react';
 import './NiceTable.css';
 
-type Row = {
+interface Row {
```
magland, line 6: There were two versions of NiceTable (in two separate places in the source tree), and so we just combined them in the extensions/common directory

src/extensions/common/NiceTable.tsx
```diff
@@ -14,9 +15,9 @@ interface Col {
 interface Props {
     rows: Row[],
     columns: Col[],
-    onDeleteRow?: ((row: Row) => void),
+    onDeleteRow?: ((key: string, columnValues: {[key: string]: any}) => void),
```
jsoules, line 18: The types passed into `onDeleteRow` and `onEditRow` look to match the `Row` interface defined above, and it looks like that had been structured that way previously--was this a conscious decision to not use that interface in this version, or did something get mixed up in combining the files?

jsoules (commented):
Seems like the overall movement here is to take widgets out of `src/components` or `src/containers` and move them into `src/etxtensions/WorkplaceView`. I'm not opposed to this, but what's the overall rule about which widgets go where?

Also we hvae to put reusable non-extension components (e.g. `NiceTable`) into `extensions/common` because of the need to support Jupyter, right? And that's the only way to make sure the directory gets copied in a consistent way?

magland:
> Seems like the overall movement here is to take widgets out of src/components or src/containers and move them into src/etxtensions/WorkplaceView. I'm not opposed to this, but what's the overall rule about which widgets go where?

Right now, anything that is shared between web version and jupyter must be inside extensions/. That's not really the intent of extensions/, but it's what we have right now. We should move to a better system, but I'm not exactly sure the best way to organized. Maybe something like:

```
shared/
    extensions/
         ext1/
         ext2/
    common/
    MountainView/
    etc.
```

src/extensions/common/NiceTable.tsx
magland: Nice catch, yeah the columnValues should not be there. I'll put this in a new issue.

PR #412, opened 2021-02-08
spike amplitudes: display y-axiseven when no units are selected

### LINKED ISSUE #380: spike amplitudes: display y-axiseven when no units are selected
Also, it'd probably be a good idea to display the y-axis and label even when no units are selected. Obviously can't do any tick marks, but it makes the axes of the graph feel more fixed/stable/consistent compared to the underlying data.

_Originally posted by @jsoules in https://github.com/flatironinstitute/labbox-ephys/issues/375#issuecomment-769893492_

magland:
Fixes #380


magland:
image

jsoules (approved):
Code's straightforward and everything checks out in the UI, good to merge

PR #413, opened 2021-02-08
Split getRecordingInfo into two properly-named files

### LINKED ISSUE #365: Split getRecordingInfo into two properly-named files
We don't strictly need two separate files, although there's a school of thought that says that that's an easier way to arrange the code--fewer merge conflicts and easier to track down what each file does and where the code should be found, at the expense of potentially more imports...

Upon reflection, given that the imports problem can be managed by using an index file, I think we probably should split this into `useRecordingInfo` and `useSortingInfo` files.

_Originally posted by @jsoules in https://github.com/laboratorybox/labbox-ephys/pull/361#discussion_r565489720_

magland:
Fixes #365


magland:
Some import paths in other branches may need to be adjusted to conform.

jsoules (approved):

PR #430, opened 2021-02-10
Cross-correlograms: selection should be independent of global state

### LINKED ISSUE #329: Cross-correlograms: selection should be independent of global state

magland:
Fixes #329


magland:
@jsoules, this is an intersting example of using a custom hook to modify the selection/selectionDispatch pair.

jsoules (approved):
This looks good.

PR #431, opened 2021-02-10
handle case where all electrodes have same coordinate, eg (0, 0)

### LINKED ISSUE #292: handle case where all electrodes have same coordinate, eg (0, 0)

magland:
Fixes #292

@jsoules. This started out as a simple fix to the degenerate case of all electrodes having the same coordinates. But then I found that ElectrodeGeometryWidget did not use the same code as the other widgets. So then it took some work to get that to use the same layers. That's why there are little adjustments to the types throughout. Nothing major.

### DIFF
--- a/src/extensions/averagewaveforms/AverageWaveformsView/setupElectrodes.ts
+++ b/src/extensions/averagewaveforms/AverageWaveformsView/setupElectrodes.ts
@@ -127,21 +127,21 @@ const setupElectrodes = (args: {width: number, height: number, electrodeLocation
     if (layoutMode === 'vertical') {
         return setupVerticalElectrodes({width, height, electrodeIds})
     }
-    const electrodeLocations2 = fixDegenerateCase(electrodeLocations)
+    const correctedElectrodeLocations = fixDegenerateCase(electrodeLocations)
     const W = width - 10 * 2
     const H = height - 10 * 2
     const canvasAspect = W / H
 
-    const radius = computeRadius(electrodeLocations2)
-    let boundingBox = getElectrodesBoundingBox(electrodeLocations2, radius)
+    const radius = computeRadius(correctedElectrodeLocations)
+    let boundingBox = getElectrodesBoundingBox(correctedElectrodeLocations, radius)
     let boxAspect = getWidth(boundingBox) / getHeight(boundingBox)
 
-    let realizedElectrodeLocations = electrodeLocations2
+    let realizedElectrodeLocations = correctedElectrodeLocations
     if ((boxAspect > 1) !== (canvasAspect > 1)) {
         // if the two aspect ratios' relationship to 1 is different, then one is portrait
         // and the other landscape. We should then correct by rotating the electrode set 90 degrees.
-        // note: a 90-degree right rotation in 2d makes x' = y and y' = -x
-        realizedElectrodeLocations = electrodeLocations2.map((loc) => {
+        // note: a 90-degree rotation around the origin makes x' = y and y' = -x
+        realizedElectrodeLocations = correctedElectrodeLocations.map((loc) => {
             return [loc[1], -loc[0]]
         })
         // and of course that also means resetting the x- and y-ranges of the bounding box.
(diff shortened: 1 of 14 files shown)


src/extensions/averagewaveforms/AverageWaveformsView/setupElectrodes.ts
```diff
@@ -102,7 +102,22 @@ const setupVerticalElectrodes = ({width, height, electrodeIds}: {width: number,
...
 
-const setupElectrodes = (args: {width: number, height: number, electrodeLocations: Vec2[], electrodeIds: number[], layoutMode: 'geom' | 'vertical'}): {
+const fixDegenerateCase = (electrodeLocations: Vec2[]) => {
+    const box = {
+        xmin: getArrayMin(electrodeLocations.map(e => (e[0]))),
+        xmax: getArrayMax(electrodeLocations.map(e => (e[0]))),
+        ymin: getArrayMin(electrodeLocations.map(e => (e[1]))),
+        ymax: getArrayMax(electrodeLocations.map(e => (e[1])))
+    }
+    if ((box.xmin === box.xmax) && (box.ymin === box.ymax)) {
```
jsoules, line 112: So this still works okay if one of the dimensions is zero? (I see it's mapped to a linear geometry, but I was also wondering what would happen if the electrodes were actually laid out in a line anyway)

src/extensions/averagewaveforms/AverageWaveformsView/setupElectrodes.ts
```diff
@@ -112,20 +127,21 @@ const setupElectrodes = (args: {width: number, height: number, electrodeLocation
...
+    const radius = computeRadius(electrodeLocations2)
+    let boundingBox = getElectrodesBoundingBox(electrodeLocations2, radius)
     let boxAspect = getWidth(boundingBox) / getHeight(boundingBox)
 
-    let realizedElectrodeLocations = electrodeLocations
+    let realizedElectrodeLocations = electrodeLocations2
     if ((boxAspect > 1) !== (canvasAspect > 1)) {
         // if the two aspect ratios' relationship to 1 is different, then one is portrait
         // and the other landscape. We should then correct by rotating the electrode set 90 degrees.
         // note: a 90-degree right rotation in 2d makes x' = y and y' = -x
```
jsoules, line 143: A 90-degree right rotation *about the origin*, I meant here.

src/extensions/averagewaveforms/AverageWaveformsView/setupElectrodes.ts
```diff
@@ -112,20 +127,21 @@ const setupElectrodes = (args: {width: number, height: number, electrodeLocation
     if (layoutMode === 'vertical') {
         return setupVerticalElectrodes({width, height, electrodeIds})
     }
+    const electrodeLocations2 = fixDegenerateCase(electrodeLocations)
```
jsoules, line 130: Might be more descriptive to call this something like 'correctedElectrodeLocations', or just have it be `let realizedElectrodeLocations` (l. 139) from the start

jsoules (commented):
As per your note, a lot of these changes are standardizing the way that electrode geometry is handled to comply with the other widgets; this is definitely a good thing.

I had a couple notes on the specific fix, but nothing urgent.

src/extensions/averagewaveforms/AverageWaveformsView/setupElectrodes.ts
magland: I don't think it's a problem if they are laid out linearly. The estimated spacing is then non-zero.

src/extensions/averagewaveforms/AverageWaveformsView/setupElectrodes.ts
magland: i'll update that

src/extensions/averagewaveforms/AverageWaveformsView/setupElectrodes.ts
magland: you're right, I'll update that

PR #432, opened 2021-02-10
jupyter compilation - give hint to avoid memory crash during compile

### LINKED ISSUE #411: jupyter compilation - give hint to avoid memory crash during compile
If run out of memory while building the jupyter extension, the following can remedy:

```
export NODE_OPTIONS="--max-old-space-size=8192"
```

magland:
Fixes #411


magland:
Just modifying the docs.

jsoules (approved):
LGTM. Noting that we should probably update the name of `example2.ipynb` if example1 is no longer part of the repo (as I think is planned?)

PR #433, opened 2021-02-11
WorkspaceView in jupyter: ask before delete?

### LINKED ISSUE #424: WorkspaceView in jupyter: ask before delete?
right now pressing the trash can button immediately deletes the recording, but maybe it should ask the user if they really want to do it first since this cannot be undone easily

magland:
Fixes #424

Add confirm delete and clean up code in the NiceTable, SortingsTable, and RecordingsTable

### DIFF
--- a/src/extensions/common/NiceTable.tsx
+++ b/src/extensions/common/NiceTable.tsx
@@ -1,6 +1,6 @@
 import { Checkbox, IconButton, Table, TableBody, TableCell, TableHead, TableRow } from '@material-ui/core';
 import { Delete, Edit } from "@material-ui/icons";
-import React, { FunctionComponent } from 'react';
+import React, { FunctionComponent, useCallback, useMemo, useState } from 'react';
 import './NiceTable.css';
 
 interface Row {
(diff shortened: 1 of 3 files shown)


magland:
image

src/extensions/common/NiceTable.tsx
```diff
@@ -109,6 +142,51 @@ const NiceTable: FunctionComponent<Props> = ({
...
+        <span>
+            Confirm delete?
+            <IconButton
+                title={title}
+                onClick={handleClick}
+            ><Delete /></IconButton>
+            <IconButton
+                title={"Cancel"}
+                onClick={handleCancel}
+            >🗙</IconButton>
```
jsoules, line 174: This is a cancel emoji. For some reason, it's not displaying correctly with my installed font set, either on github or in vscode or (rather more importantly) through the web ui. To ensure consistency across users' font sets, we should probably encode it with a `FontAwesome`/`Material`/etc embedded font icon as we do elsewhere.

jsoules (changes requested):
Changes make sense and work for me in the UI, but I've left a note about the cancel icon not displaying correctly--looks like there's a raw emoji in the source code, which isn't playing nicely with my installed font set.

src/extensions/common/NiceTable.tsx
magland: I have replaced it by `&#10006;` which should do the trick.

jsoules: Confirming it looks good now.

PR #447, opened 2021-02-16
Separate out application-unspecific labbox functionality

magland:


magland:
Merging with all my fingers and toes crossed.

